PR #32, opened 2014-12-18
Add --bump, --tag, and --tag-pattern options

raphink:
This PR uses [`puppet-blacksmith`](https://github.com/maestrodev/puppet-blacksmith) to automate the following tasks:

* Bump minor version in `metadata.json` if files were changed
* Tag module with new version (using `v%s` by default, can be overridden with --tag-pattern)

This allows for bulk minor releases using msync.

### DIFF
--- a/lib/modulesync/cli.rb
+++ b/lib/modulesync/cli.rb
@@ -14,6 +14,7 @@ module ModuleSync
         :git_provider_address => 'github.com',
         :managed_modules_conf => 'managed_modules.yml',
         :configs              => '.',
+        :tag_pattern          => '%s',
       }
     end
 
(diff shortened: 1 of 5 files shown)


raphink:
@cmurphy, @mcanevet  Rebased.

lib/modulesync/cli.rb
```diff
@@ -35,7 +36,7 @@ def parse_opts(args)
       @options.merge!(Hash.transform_keys_to_symbols(Util.parse_config(MODULESYNC_CONF_FILE)))
       @options[:command] = args[0] if commands_available.include?(args[0])
       opt_parser = OptionParser.new do |opts|
-        opts.banner = "Usage: msync update [-m <commit message>] [-c <directory> ] [--noop] [-n <namespace>] [-b <branch>] [-f <filter>] | hook activate|deactivate [-c <directory> ] [-n <namespace>] [-b <branch>]"
+        opts.banner = "Usage: msync update [-m <commit message>] [-c <directory> ] [--noop] [--bump] [--tag] [--tag-pattern <tag_pattern] [-n <namespace>] [-b <branch>] [-f <filter>] | hook activate|deactivate [-c <directory> ] [-n <namespace>] [-b <branch>]"
```
cmurphy: I think there's a typo here. there should be an extra '>' after tag_pattern.

raphink: Fixed.

lib/modulesync/git.rb
```diff
@@ -27,9 +28,25 @@ def self.pull(git_base, name, branch, opts)
       end
     end
 
+    def self.bump(repo, m)
+      m.bump!
+      puts "Bumped to version #{m.version}"
```
cmurphy: This is reporting the old version instead of the new version, which is confusing when looking at the output.

It looks like puppet-blacksmith returns the new version from `bump!`, could we instead save new_version as a variable and report that instead of m.version?

lib/modulesync/git.rb
```diff
@@ -27,9 +28,25 @@ def self.pull(git_base, name, branch, opts)
...
+      repo.add('metadata.json')
+      repo.commit("Release version #{m.version}")
```
cmurphy: Also problematic here, because the metadata version is different than what the commit indicates.

cmurphy:
Sorry for leaving this for so long. I have one more issue that I think needs to be resolved. The bump method is problematic because blacksmith is reporting the version as the old version instead of the new one. This means the output is confusing, the commit message is wrong, and tagging will usually fail because the tag will already exist.

cmurphy:
Could you also add documentation to the readme?

raphink:
I fixed the issue and added some documentation.

PR #33, opened 2014-12-18
Add support for updating CHANGELOG.md

raphink:
Depends on #32


raphink:
@cmurphy, @mcanevet rebased on new #32

cmurphy:
Could you add documentation to the readme? I was surprised at the behavior of only adding the most recent change to the changelog. It makes sense given your usecase from #32 but it should be documented.

raphink:
Updated and added some doc.

PR #34, opened 2014-12-22
Fix support for branches other than 'master'.

razorsedge:
This seems to better support those of us who do not commit directly to the master branch.  I was having a terrible time with the code making a new branch instead of switching to the existing remote.  `repo.checkout(branch)` was the magic sauce.

### DIFF
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -55,7 +55,7 @@ module ModuleSync
 
       managed_modules.each do |puppet_module|
         puts "Syncing #{puppet_module}"
-        Git.pull(options[:git_user], options[:git_provider_address], options[:namespace], puppet_module)
+        Git.pull(options[:git_user], options[:git_provider_address], options[:namespace], options[:branch], puppet_module)
         module_configs = Util.parse_config("#{PROJ_ROOT}/#{puppet_module}/#{MODULE_CONF_FILE}")
         files_to_manage = module_files | defaults.keys | module_configs.keys
         files_to_delete = []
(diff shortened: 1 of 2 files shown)


cmurphy:
Could you rebase against master? 03db55a does a lot of similar work and is causing conflicts.

I'm seeing issues in master and with this change where git pull won't work if the branch doesn't exist in the remote. Could we think about (probably in a future change) having some option to fetch from origin/master and checkout a new branch based on that?

cmurphy:
I'm not sure why `repo.branch(branch).checkout is a problem. According to the git gem documentation it should work fine on an existing local branch, and after trying it out I'm not able to find an issue with it:

```ruby
> repo = Git.open('.')
 => #<Git::Base:0x007ff2e511ff28 @logger=nil, @working_directory=#<Git::WorkingDirectory:0x007ff2e511fcd0 @path="/private/tmp/modulesync">, @repository=#<Git::Repository:0x007ff2e511fb90 @path="/private/tmp/modulesync/.git">, @index=#<Git::Index:0x007ff2e511faf0 @path="/private/tmp/modulesync/.git/index">>
> repo.branches.collect { |r| r.name }
 => ["master", "HEAD -> origin/master", "master"]
> repo.branch("newbranch").checkout
 => "Switched to branch 'newbranch'"
> repo.branches.collect { |r| r.name }
 => ["master", "newbranch", "HEAD -> origin/master", "master"]
> repo.branch("newbranch").checkout
 => "Already on 'newbranch'"
```

Modulesync seems to work fine as long as the branch exists on the remote.

razorsedge:
I hope to have some time to rebase this weekend.

It was 7 weeks ago that I last worked on this, so I am having trouble remembering, but I think the problem arises with `repo.reset_hard` doing things I did not expect to the local repository.  `repo.checkout(branch)` was the only way I could get it to work.

PR #36, opened 2015-02-06
Add a global config key

DavidS:
Using the new :global entry in config_defaults.yml and .sync.yml,
data can be shared across files in a module.

### DIFF
--- a/lib/modulesync/constants.rb
+++ b/lib/modulesync/constants.rb
@@ -6,5 +6,6 @@ module ModuleSync
     MODULESYNC_CONF_FILE = 'modulesync.yml'
     PROJ_ROOT            = './modules'
     HOOK_FILE            = '.git/hooks/pre-push'
+    GLOBAL_DEFAULTS_KEY  = :global
   end
 end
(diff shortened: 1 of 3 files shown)


lib/modulesync/constants.rb
```diff
@@ -6,5 +6,6 @@ module Constants
     MODULESYNC_CONF_FILE = 'modulesync.yml'
     PROJ_ROOT            = './modules'
     HOOK_FILE            = '.git/hooks/pre-push'
+    GLOBAL_DEFAULTS_KEY  = :global
```
cmurphy: This should probably be 'global' instead of :global so that users don't have to use a ruby symbol in the yaml file.

DavidS: having a plain string there avoids colliding with an actual file named 'global'. The yaml would look like this:

```
---
:global:
  key: value
```

Which is e.g. also used by the foreman in its configurations, and I do not think it to be too bad.

DavidS: Pondering collisions more, It might also make sense to NOT merge the global section into `@configs` and put it into a separate member of `ForgeModuleFile`.

cmurphy:
I think using the :global key in @configs is fine. Did you have any further thoughts on this?

DavidS:
Hi, if you're OK with:global, I'd ask you to merge it as is. I've rebased to current master to resolve the conflict.

PR #38, opened 2015-02-10
Add support for Gerrit review system

Spredzy:
This commits add supports gerrit system and will submit a review
everytime a change is made on the same branch.

The `modulesync.yml` file looks like

```
---                                                                                                                                                              
namespace: stackforge
git_base:  ssh://jdoe@review.openstack.org:29418/
branch: msync_foo
branch_prefix: HEAD:refs/publish/master/
pre_commit_script: openstack-commit-msg-hook.sh

```

When a user receive a -1, the user can simply re-run `msync update --amend` and a new patchset will be submited

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,18 @@ found.
 msync update -m "Commit message"
 ```
 
+If the user decide to use a different branch than master, s/he might want to amend the commit based on users feedbacks.
+
+```
+msync update --amend
+```
+
+If a push --force is needed.
+
+```
+msync update --amend --force
+```
+
 #### Automating Updates
 
 You can install a pre-push git hook to automatically clone, update, and push
@@ -199,6 +211,35 @@ Then you can run ModuleSync without extra arguments:
 ```
 msync update --noop
 msync update -m "Commit message"
+```
+
+Available parameters for modulesync.yml
+
+* git_base : The default URL to git clone from (Default: 'git@github.com:')
+* namespace : Namespace of the projects to manage (Default: 'puppetlabs')
+* branch : Branch to push to (Default: 'master')
+* remote_branch : Remote branch to push to (Default: Same value as branch)
+* pre_commit_script : A script to be run before commiting (ie. contrib/myfooscript.sh)
+
+##### Example
+
+###### Github
+
+```
+---
+namespace: MySuperOrganization
+branch: modulesyncbranch
+```
+
+###### Gerrit
+
+```
+---
+namespace: stackforge
+git_base:  ssh://jdoe@review.openstack.org:29418/
+branch: msync_foo
+remote_branch: refs/publish/master/msync_foo
+pre_commit_script: openstack-commit-msg-hook.sh
 ```
 
 #### Filtering Repositories
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -56,11 +56,7 @@ module ModuleSync
       # managed_modules is either an array or a hash
       managed_modules.each do |puppet_module, opts|
         puts "Syncing #{puppet_module}"
-        if options[:git_base]
-          git_base = options[:git_base]
-        else
-          git_base = "#{options[:git_user]}@#{options[:git_provider_address]}:#{options[:namespace]}"
-        end
+        git_base = "#{options[:git_base]}#{options[:namespace]}"
         Git.pull(git_base, puppet_module, options[:branch], opts || {})
         module_configs = Util.parse_config("#{PROJ_ROOT}/#{puppet_module}/#{MODULE_CONF_FILE}")
         files_to_manage = module_files | defaults.keys | module_configs.keys
@@ -80,9 +76,9 @@ module ModuleSync
         end
         files_to_manage -= files_to_delete
         if options[:noop]
-          Git.update_noop(puppet_module, options[:branch])
+          Git.update_noop(puppet_module, options)
         else
-          Git.update(puppet_module, files_to_manage, options[:message], options[:branch], options[:bump], options[:changelog], options[:tag], options[:tag_pattern])
+          Git.update(puppet_module, files_to_manage, options)
         end
       end
     elsif options[:command] == 'hook'
--- a/lib/modulesync/git.rb
+++ b/lib/modulesync/git.rb
@@ -64,10 +64,16 @@ module ModuleSync
     end
 
     # Git add/rm, git commit, git push
-    def self.update(name, files, message, branch, bump=false, changelog=false, tag=false, tag_pattern=nil)
+    def self.update(name, files, options)
       module_root = "#{PROJ_ROOT}/#{name}"
+      message = options[:message]
+      if options[:remote_branch]
+        branch = "#{options[:branch]}:#{options[:remote_branch]}"
+      else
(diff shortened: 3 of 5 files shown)


lib/modulesync.rb
```diff
@@ -82,7 +91,7 @@ def self.run(args)
         if options[:noop]
           Git.update_noop(puppet_module, options[:branch])
         else
-          Git.update(puppet_module, files_to_manage, options[:message], options[:branch])
+          Git.update(puppet_module, files_to_manage, options[:message], options[:branch], options[:git_provider], options[:git_user], options[:git_provider_address], options[:git_provider_port])
```
dvorak: Probably should wrap this line.

dvorak:
Thanks for working on this!  Other than a few long lines, it looks ok to me, but I'll defer to others with more experience on modulesync internals.

Spredzy:
@cmurphy would you agree that I pass the options hash instead of passing each items separately? cc @dvorak

lib/modulesync/git.rb
```diff
@@ -39,7 +41,20 @@ def self.update(name, files, message, branch)
...
+        case git_provider
+        when 'gerrit'
+          branch = "HEAD:refs/publish/master/#{branch}"
+          if ! File.exists? ("#{PROJ_ROOT}/#{name}/.git/hooks/commit-msg")
+            Dir.chdir("#{PROJ_ROOT}/#{name}") {
+              %x[gitdir=$(git rev-parse --git-dir); scp -p -P #{git_provider_port} #{git_user}@#{git_provider_address}:hooks/commit-msg ${gitdir}/hooks/]
+            }
+          else
+            opts = {:amend => true}
+            message = nil
```
cmurphy: These two lines are extremely problematic and unnecessary. They cause the script to amend the latest commit rather than making a new commit, when the previous commit is likely to be a change already in master and not owned by the committer.

lib/modulesync/git.rb
```diff
@@ -19,16 +19,18 @@ def self.pull(git_base, name, branch, opts)
 
       # Repo already cloned, check out master and override local changes
       else
-        puts "Overriding any local changes to repositories in #{PROJ_ROOT}"
-        repo = ::Git.open("#{PROJ_ROOT}/#{name}")
-        repo.branch(branch).checkout
-        repo.reset_hard
-        repo.pull('origin', branch)
+        if ! git_provider.eql? 'gerrit'
```
cmurphy: The local repo should start with an updated state regardless of the git provider.

lib/modulesync/git.rb
```diff
@@ -19,16 +19,18 @@ def self.pull(git_base, name, branch, opts)
...
+          repo.branch(branch).checkout
+          repo.reset_hard
+          repo.pull('origin', branch)
+        end
       end
     end
 
     # Git add/rm, git commit, git push
-    def self.update(name, files, message, branch)
+    def self.update(name, files, message, branch, git_provider, git_user, git_provider_address, git_provider_port)
```
cmurphy: This won't work if the user's supplied the full git_base instead of the individual git_* parameters.

cmurphy:
I'm a little frustrated that gerrit must be treated so differently...but looking at your changes and playing with git-review a bit, I think perhaps it doesn't have to be treated so differently.

I propose that instead of a 'git_provider' parameter, we instead just need a 'pre_commit_script' parameter. We can maybe have a 'contrib/' or 'scripts/' directory that contains something to check for the existence of and install the commit-msg hook. The user can supply a gerrit-style git_base parameter already, as well as supply 'refs/publish/master/branchname' as the branch name. We can add documentation to make this easier. There isn't any need, as far as I can see, to explicitly amend commits, as the pre-commit hook will do that automatically.

What do you think?

Spredzy:
@cmurphy this PR has been reworked with the following changes : 

* Unification of parameter: Only the `git_base` parameter is now used, default being `git@github.com:` (no more `git_user`, `git_provider_address`)
* Creation of the contrib/ folder: As suggested using a `contrib/` folder for pre_commit_script. In the case of review.openstack.org, it will bring the commit-msg hook directly from review.openstack.org (ie. `openstack-commit-msg-hook.sh`)
* New command available `--amend` and `--amend --force` : In a system of review (like gerrit) msync update --amend will be enough to create a new patchset, in a system like github, the user will have to explicitly specify `--amend --force`

I hope all the point abode have been addressed

dvorak:
Would it make sense to just remove the --amend option entirely and have the pre-commit script check if the change already includes a Change-ID and add it (amend) if it's not there?  If someone is using Gerrit, then that's very likely the behavior they want, and they can modify the pre-commit script if not.

README.md
```diff
@@ -201,6 +213,35 @@ msync update --noop
 msync update -m "Commit message"
 ```
 
+Available parameters for modulesync.yml
+
+* git_base : The default URL to git clone from (Default: 'git@github.com:')
+* namespace : Namespace of the projects to manage (Default: 'puppetlabs')
+* branch : Branch to push to (Default: 'master')
+* branch_prefix : Branch prefix if necessary (ie. HEAD:refs/publish/master/)
```
cmurphy: Instead of `branch` and `branch_prefix` could we have `branch` (which would be the local branch) and `remote_branch` (which would default to the value of `branch`)? That way the push looks like `git push origin localbranchname:remotebranchname`, so they could potentially have different names and we don't need to refer to `HEAD:`.

lib/modulesync.rb
```diff
@@ -82,7 +78,7 @@ def self.run(args)
         if options[:noop]
           Git.update_noop(puppet_module, options[:branch])
         else
-          Git.update(puppet_module, files_to_manage, options[:message], options[:branch], options[:bump], options[:changelog], options[:tag], options[:tag_pattern])
+          Git.update(puppet_module, files_to_manage, options[:message], options[:branch], options[:branch_prefix], options[:bump], options[:changelog], options[:tag], options[:tag_pattern], options[:pre_commit_script], options[:amend], options[:force])
```
cmurphy: In an earlier comment you asked about passing the whole options hash instead of each value...I +1 that idea, this is starting to look silly. :)

cmurphy:
@Spredzy looking good! A couple more notes.

@dvorak I'd prefer not to have --amend be implicit, as it has the potential to be destructive. Moreover, as much as possible, I'd like to not have to treat gerrit or any other git providers as special. Adding the --amend and --force options make them available to github/gitlab/gitolite users. In any case, if you were doing it by hand you would use --amend anyway, so I don't think it's a big deal to require it here.

lib/modulesync/git.rb
```diff
@@ -76,8 +77,22 @@ def self.update(name, files, message, branch, bump=false, changelog=false, tag=f
...
       end
       begin
-        repo.commit(message)
-        repo.push('origin', branch)
+        branch = "#{branch_prefix}#{branch}"
+        opts_commit = {}
+        opts_push = {}
+        if amend
+          opts_commit = {:amend => true}
+          message = nil
```
cmurphy: Do we need to set this to nil? I've had a lot of cases where amending the commit also means changing the commit message.

PR #43, opened 2015-03-10
Pass puppet_module to ERB

raphink:
This allows to access the module name using
the @module variable.

### DIFF
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -63,6 +63,7 @@ module ModuleSync
         files_to_delete = []
         files_to_manage.each do |file|
           file_configs = (defaults[file] || {}).merge(module_configs[file] || {})
+          file_configs[:puppet_module] = puppet_module
           if file_configs['unmanaged']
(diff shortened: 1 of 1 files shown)


raphink:
@cmurphy this PR would be very useful to generate docs for example (with links to issues URL, etc.).

cmurphy:
@raphink instead of adding a new instance variable to the template, could :puppet_module be added as a key in the file_configs variable? Then it would be wrapped in the @configs variable.

raphink:
@cmurphy I wanted to avoid that as that might conflict with existing parameters in @configs.

cmurphy:
@raphink if there's already a :puppet_module key in @configs then the user wouldn't make use of the @module variable anyway.

raphink:
Updated the commit. Do you like this better @cmurphy ?

cmurphy:
:+1:  thanks!

raphink:
@cmurphy thanks. Could you cut a new release, please?

cmurphy:
@raphink yes, will do this week.

PR #47, opened 2015-06-30
Allow one to sync non-bare local git repository

Spredzy:
One can currently manage bare local git repositories with modulesync but
cannot manage non-bare git repositories. This commit aims to add this
feature.

If I was to have several local directories (from git clone), I can't keep
them in sync with modulesync as git clone would complain about :

fatal: '/tmp/puppet/openstack/puppet-nova.git' does not
appear to be a git repository
fatal: Could not read from remote repository.

Yet '/tmp/puppet/openstack/puppet-nova' is a valid path with a .git
directory in it.

This is due to the '.git' appended at the end when protocol is file.


sbadia:
:+1:

PR #50, opened 2015-07-08
Adding a few tests

danzilio:


danzilio:
New PR. Looks like the tests ran fine. I'll submit a PR later this week with some more tests and some refactored code!

PR #51, opened 2015-07-23
This is basically a complete rewrite of ModuleSync.

danzilio:
I removed some features like the ability to update a changelog, or to bump and tag modules. These are things that should be done outside of ModuleSync. We should stick to managing modules. I'm open to adding some of this functionality back if it's absolutely necessary but it would come with lots of code smell...


danzilio:
I've also got a few items on my personal todo list:
  - Better logging
  - More tests
  - Better docs

I just wanted to get this out there. I've been working on this for a bit (I've gone through a few iterations); figured it wasn't doing anybody any good by not submitting it.

ferventcoder:
I personally feel that modulesync is not the proper name for what this tool does - it's more like a git/git files sync tool. That said, some of the things you mentioned removing seem like things that would be more for managing modules and not just a files sync.

danzilio:
I got the same sense mucking around with this...there has to be a bigger use case we can fulfill...

Generic boilerplate management?

ferventcoder:
I'm still new to modsync - the only thing I can offer here is git advice - this is a huge refactoring to be done in one commit. If the maintainers feel this is a direction they want to go, consider breaking up the changes into manageable chunks. And the second is each commit should explain the change it makes and why the change is good and/or necessary.  A good commit message has a 50 character or less summary, followed by an empty line, and a body that goes into detail explaining the change.

danzilio:
Yeah, this is definitely more than just refactoring. This definitely changes some external behaviors. I definitely understand that this is a lot but I'm willing to work to get this merged. I've written some tests, but it's hard to do something like this incrementally without a baseline.

danzilio:
Also, I'm not sure why I did this on master, I always use a feature branch!

danzilio:
Resubmitting on a feature branch.

PR #52, opened 2015-07-23
WIP: Redesign

danzilio:
I removed some features like the ability to update a changelog, or to bump and tag modules. These are things that should be done outside of ModuleSync. We should stick to managing modules. I'm open to adding some of this functionality back if it's absolutely necessary but it would come with lots of code smell...

Resubmitting #51 as this could take a while to get merged in.


danzilio:
I probably should've been a little more clear about the intent of this. I don't expect this PR to be merged without some serious discussion. I've completely changed the way we approach some things and this is meant to initiate a dialogue with the maintainers about how I've approached some of these problems. I initially dug into this because I wanted to write some rake tasks for ModuleSync, but when I dug into the code I realized that there was really no API outside of the command line client. I've broken a lot of the functionality out into classes and made it a lot more modular. In the process of doing this I pulled out some features that didn't seem to 'fit' and seemed to detract from the overall maintainability of ModuleSync. I've also written some tests. They're not perfect, but the code coverage is over 99%. This should help when extending ModuleSync in the future.

I was also experimenting while I was doing this. I went through a few iterations and ended up here. There's still plenty of work to do. I just wanted to get the code out there so I could see what people thought about these changes. The code wasn't doing anybody any good just sitting on my laptop.

That being said, what do people think?

cmurphy:
014706c21a0e83d45f5eefb6f4c19e781fc91007 is far too big a commit to review on its own, it would be easier to review if it was broken up into small chunks, with each commit message explaining the purpose.

This tool was designed specifically to help manage puppet modules, not just as a git sync tool as @ferventcoder suggested in #51 . @raphink did a lot of work to add features that you removed, so I expect he might have feelings about this change. @Spredzy is leading the effort to integrate this tool with the OpenStack puppet modules and depends highly on the --force and --amend options that were added back after the initial commit of this PR.

In my mind this is not a complex tool. It was meant to fulfill a rather narrow purpose and does that pretty well. I'm curious, if you thought it deserved a complete rewrite then why not just write a different tool to fulfill the purpose you need?

ferventcoder:
@cmurphy of course I also mentioned below that of my new-ness to modulesync and that the bits being removed seemed akin to what would make this a modulesync instead of just a git files sync. Rereading my comment, I see that it's confusing to understand the purpose of my comment, which was originally to point out that this is a tool for modules syncing and it appeared that the bits specific to that looked like they were being removed.

ferventcoder:
I wasn't trying to say whether the changes were good or not, that's your call and not mine. :)

danzilio:
Thanks for the feedback @cmurphy! Let me preface this by saying how much I love ModuleSync. It's made my life so much easier.

I didn't set out to completely rewrite this, it happened almost accidentally. I initially set out only to extend ModuleSync. In my effort to extend ModuleSync I set out to write some tests. I quickly realized that ModuleSync wasn't written with testability in mind (this is not an indictment, just an observation). Accepting contributions is much easier when there's good test coverage because it helps maintainers understand the scope and quality of a contribution. Making a contribution is also much easier with tests because it allows the contributor to validate their changes against an existing corpus of knowledge. Tests also signal the intent of the developer, which allows contributors to make changes that are compatible with the intent of the creators. I felt that I couldn't confidently make a significant contribution without writing some tests first. Given the lack of tests, and the existing design of the code, I was faced with a decision. Do I make a relatively complex contribution without tests, or should I take a more holistic approach and attempt to make the code more testable (and, consequently, more modular, easy to extend, etc)? I opted to try and make the code more modular and more easily tested.

This resulted in some seriously major changes. Given how tightly coupled the code was when I started out, I'm not entirely sure how I could've done this incrementally enough to satisfy anybody. Almost all of the tool's logic was encapsulated inside in the `ModuleSync` module's `run` method. Incremental changes to that method rippled out quite a bit and necessitated significant changes elsewhere in the code. I ultimately felt that taking a retrofit approach to this would be a disservice to the ModuleSync community.

Now, as I became more familiar with the code, I realized that there were some features that were particularly un-maintainable. I started to think about more maintainable approaches to this functionality. As you can probably see, I've added back most of the features that I removed in my initial contribution (with some modifications). I'm not entirely sure why we need to include the OpenStack Project's hook scripts in this gem as the OpenStack project is just one stakeholder of this project. The way this was written, If I wanted to use a different hook script, I'd have to submit a PR for this project and wait for a new release. IMHO, this is a rather silly approach. I added this functionality back, but with the ability to configure any hook script relative to the project root.

I'm also not sure why we would want to update a changelog as there's no accepted standard format for changelogs. The closest thing would be http://keepachangelog.org, but this tool didn't even support that format (it also didn't support a configurable format, templating, or anything other than what was hardcoded). My changelogs probably don't look like your changelogs ;). The same goes for bumping and tagging modules. There are many different use cases that these features didn't take into account. These features may be in use by some people, but they're not particularly maintainable features. Moreover, I think that this redesign of ModuleSync enables this functionality to be externalized from this tool. A user could update the changelog as well as bump and tag a module version in a hook script.

I'm not particularly interested in making this a generic git syncing tool, although I'll admit there's definitely a use case there. I also agree with you, this is not a particularly complex tool. There was a lot of complexity in certain places that made it very risky to make changes. I wanted to address some of those things because I'm absolutely crazy about this tool! I don't think this PR introduces unnecessary complexity; I actually think this simplifies things quite a bit. I'll also acknowledge that there may be better ways to do this. That's why I submitted the PR; to have a discussion. I don't have all the answers and I'm definitely not the best developer out there. I was hoping to get some ideas from the community about how to approach these problems.

Please don't take this as an attempt to co-opt the project or as an indictment of the code. I really love this tool and I want to help make it even more fantastic. I'd love to hear what you think about my thought process and approach. I'm certainly willing to put the effort in to make these contributions acceptable you and to the community at large.

cmurphy:
I am all for improving modularity and testability. Indeed this was not written with testing in mind and anything we can do to fix that is a positive step.

Since this rewrite is so drastic it is quite difficult to review. I would like to see the commits reorganized so that we can see step by step how things are breaking out into new modules. If possible I would like to avoid commits that add back functionality and instead just leave or fix the functionality.

I agree that the changelog feature is not quite how I generally handle changelogs, but it seemed like a useful feature for some. I'd like to hear from @raphink on how to handle that.

danzilio:
@cmurphy I can definitely do that!

raphink:
Pfiew, that's quite a backlog to read!

First, I would agree with @ferventcoder in that I see modulesync as a generic tool. I've actually been tempted in the past to get rid of `module` variables in the code as they narrowed the scope of the tool imo.

I've been using modulesync for two types of projects: syncing Puppet modules (for Camptocamp, Augeasproviders.com and Puppet-community's puppet-lint plugins) as well as our R10K repositories. That's right: At Camptocamp, we manage 4 separate infrastructures that share modules and data, and we want to use the same versions of Puppet modules on each of them. So, we generate our R10K repos from a central template repository and deploy them using modulesync. That's enough for me to consider modulesync as a generic tool, that could be used for any similar need, Puppet-related or not.

As for testing, I totally agree with @danzilio. I strongly regret the lack of tests on this piece of software. It's made features hard to add in the past, without breaking current functionalities.

When it comes to the `bump`, `tag`, and `changelog` features, they come from our use of modulesync in a Continuous Deployment Pipeline system for our Puppet modules. We use Travis CI and automatically deploy modules to the Puppet Forge when tagged builds pass. In some cases (e.g. adding a puppet-lint plugin), it's been very useful for us to be able to sync modules, tag them with a minor revision, update the changelog with the msync message, and push, as this will automatically release all modules whose build passes with the new plugin. I think ideally, modulesync should be more… modular, with the possibility of adding such features as plugins.

As for standards in changelogs, this is indeed a common issue. I've personnally started looking at [`github_changelog_generator`](https://github.com/skywinder/github-changelog-generator), which follows the "standard" from http://keepachangelog.com. I've already used it in [one Puppet module](https://forge.puppetlabs.com/camptocamp/augeas_file/changelog) and would welcome a way to plug that as a Rake task for example.

danzilio:
Thanks for your input @raphink! I appreciate hearing more use cases!

I originally set out on this journey because I didn't want to maintain a `managed_modules.yml` file. I wanted to be able to query some data source (for us, that would be Jenkins) for the list of modules to manage. I set out to write a rake task to query the Jenkins API for a list of projects in a certain folder and then pass that list to ModuleSync.

I've also been using this as a replacement for the `puppet module generate` tool (I don't think I'm alone in this).

I'd be curious to know how you'd envision a plugin system working. I've added support for arbitrary pre and post hook scripts that execute before and after ModuleSync commits its changes. Do you think you could achieve the `bump`, `tag`, and `changelog` features that way? I quite like how `puppet-lint` handles plugins, though I'm not sure we need something that robust.

In the meantime, I'm going to focus on making this PR more coherent and more easily reviewed. I'd appreciate any feedback along the way!

danzilio:
Hopefully this PR is a little easier to follow now!

igalic:
@danzilio you still working on this?

danzilio:
Depends! Do we think it's worth continuing with this? What do we think is the best/most palatable way to proceed? I'm open to anything.

danzilio:
Spoke with @cmurphy today. I'm going to close this PR. I'm going to submit a separate PR for some of these features and we can target a 1.X release hopefully later this week. The breaking changes from this PR will be submitted in a separate PR targeting a 2.X release.

PR #53, opened 2015-08-07
Create --offline flag to disable git functionality

pabelanger:
This allows existing cloned projects to be used for testing purposes.

Signed-off-by: Paul Belanger <pabelanger@redhat.com>

### DIFF
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -56,8 +56,10 @@ module ModuleSync
       # managed_modules is either an array or a hash
       managed_modules.each do |puppet_module, opts|
         puts "Syncing #{puppet_module}"
-        git_base = "#{options[:git_base]}#{options[:namespace]}"
-        Git.pull(git_base, puppet_module, options[:branch], opts || {})
+        unless options[:offline]
+          git_base = "#{options[:git_base]}#{options[:namespace]}"
+          Git.pull(git_base, puppet_module, options[:branch], opts || {})
+        end
         module_configs = Util.parse_config("#{PROJ_ROOT}/#{puppet_module}/#{MODULE_CONF_FILE}")
         global_defaults = defaults[GLOBAL_DEFAULTS_KEY] || {}
         module_defaults = module_configs[GLOBAL_DEFAULTS_KEY] || {}
@@ -80,7 +82,7 @@ module ModuleSync
         files_to_manage -= files_to_delete
         if options[:noop]
           Git.update_noop(puppet_module, options)
-        else
+        elsif not options[:offline]
           Git.update(puppet_module, files_to_manage, options)
         end
       end
--- a/README.md
+++ b/README.md
@@ -121,6 +121,17 @@ overwrites local changes.
 msync update --noop
 ```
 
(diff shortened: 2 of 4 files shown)


igalic:
@pabelanger would you mind fixing the broken tests?

cmurphy:
There will often be cases when you do want to clone the repos, you just don't want to push. Could we instead add a different flag to handle this?

pabelanger:
@cmurphy suggestions on flag name?  --disable-git? --without-git?

cmurphy:
--no-pull? --no-clone? --without-clone?

Maybe a --offline that neither pulls nor pushes?

pabelanger:
--offline works. Let me refactor

cmurphy:
Also @danzilio might want to take a look since he's working on a refactor

lib/modulesync.rb
```diff
@@ -56,8 +56,10 @@ def self.run(args)
       # managed_modules is either an array or a hash
       managed_modules.each do |puppet_module, opts|
         puts "Syncing #{puppet_module}"
-        git_base = "#{options[:git_base]}#{options[:namespace]}"
-        Git.pull(git_base, puppet_module, options[:branch], opts || {})
+        if not options[:offline]
```
danzilio: a more idiomatic way to put this would be `unless options[:offline]`

lib/modulesync.rb
```diff
@@ -78,6 +80,9 @@ def self.run(args)
           end
         end
         files_to_manage -= files_to_delete
+        if options[:offline]
```
danzilio: also, this could be a one-liner

```
next if options[:offline]
```

danzilio:
The `--offline` option name makes sense to me! I agree with @cmurphy that this may not be desirable as the default behavior for `--noop`

rnelson0:
Could some instructions for the new flag be added to https://github.com/puppet-community/modulesync#dry-run please?

lib/modulesync.rb
```diff
@@ -78,6 +80,7 @@ def self.run(args)
           end
         end
         files_to_manage -= files_to_delete
+        next if options[:offline]
```
cmurphy: I think we do still want the update_noop to happen in offline mode (it just prints the diffs to stdout)

lib/modulesync.rb
```diff
@@ -80,7 +82,7 @@ def self.run(args)
         files_to_manage -= files_to_delete
         if options[:noop]
           Git.update_noop(puppet_module, options)
-        else
+        elsif not options[:offline]
```
cmurphy: I don't think this addresses https://github.com/puppet-community/modulesync/pull/53#discussion_r36565658

I think line 83 should be if options[:noop] || options[:offline] since we want the same behavior (printing out the changes without making updates)

pabelanger: So, if we do that, we're changing the way --noop actually works.  Since now, we get output with both --noop or --offline.

This way, you have to pass both --offline --noop if you want to see the diff.  Other wise, --offline does nothing related to git.

PR #54, opened 2015-08-07
Add --project-root flag

pabelanger:
This allows the user to now define the previously hardcoded './modules' folder.

Signed-off-by: Paul Belanger <pabelanger@redhat.com>

### DIFF
--- a/features/update.feature
+++ b/features/update.feature
@@ -30,6 +30,35 @@ Feature: update
     Given I run `cat modules/puppet-test/test`
     Then the output should contain "aruba"
 
+  Scenario: Adding a new file into foobar project-root
+    Given a file named "managed_modules.yml" with:
+      """
+      ---
+        - puppet-test
+      """
+    And a file named "modulesync.yml" with:
+      """
+      ---
+        namespace: maestrodev
+        git_base: https://github.com/
+      """
+    And a file named "config_defaults.yml" with:
+      """
+      ---
+      test:
+        name: aruba
+      """
+    And a directory named "moduleroot"
+    And a file named "moduleroot/test" with:
+      """
+      <%= @configs['name'] %>
+      """
+    When I run `msync update --noop --project-root=foobar`
+    Then the exit status should be 0
+    And the output should match /Files added:\s+test/
+    Given I run `cat foobar/puppet-test/test`
+    Then the output should contain "aruba"
+
   Scenario: Modifying an existing file
     Given a file named "managed_modules.yml" with:
       """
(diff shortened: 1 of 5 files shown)


features/update.feature
```diff
@@ -24,10 +24,10 @@ Feature: update
       """
       <%= @configs['name'] %>
       """
-    When I run `msync update --noop`
+    When I run `msync update --noop --project_root=foobar`
```
cmurphy: Should this be a separate test?

lib/modulesync.rb
```diff
@@ -14,7 +14,7 @@ def self.local_file(config_path, file)
   end
 
   def self.module_file(puppet_module, file)
-    "#{PROJ_ROOT}/#{puppet_module}/#{file}"
+    "#{puppet_module}/#{file}"
```
cmurphy: Could just add project_root as another argument and avoid having to do the string concatenation later

daenney:
To keep this in line with `--remote-branch` and `--tag-pattern` this should be `--project-root`, `-` instead of `_`. This is also the more used pattern in CLI designs for multi-word arguments.

lib/modulesync/cli.rb
```diff
@@ -35,7 +36,7 @@ def parse_opts(args)
       @options.merge!(Hash.transform_keys_to_symbols(Util.parse_config(MODULESYNC_CONF_FILE)))
       @options[:command] = args[0] if commands_available.include?(args[0])
       opt_parser = OptionParser.new do |opts|
-        opts.banner = "Usage: msync update [-m <commit message>] [-c <directory> ] [--offline] [--noop] [--bump] [--changelog] [--tag] [--tag-pattern <tag_pattern>] [-n <namespace>] [-b <branch>] [-r <branch>] [-f <filter>] | hook activate|deactivate [-c <directory> ] [-n <namespace>] [-b <branch>]"
+        opts.banner = "Usage: msync update [-m <commit message>] [-c <directory> ] [--offline] [--noop] [--bump] [--changelog] [--tag] [--tag-pattern <tag_pattern>] [-p <project_root> [-n <namespace>] [-b <branch>] [-r <branch>] [-f <filter>] | hook activate|deactivate [-c <directory> ] [-n <namespace>] [-b <branch>]"
```
daenney: Missed an `_` :smile:

pabelanger: Actually, this is correct. Look to the previous tag_pattern syntax.   Well, correct because this is how tag_pattern works :)

PR #56, opened 2015-08-12
Fix bug in git pull function

### LINKED ISSUE #55: msync git clone with 'path already exists'
When attempting to run msync, the first run works as expected. 
However any further runs fail, as shown here: https://gist.github.com/fatmcgav/98629849f717c1dfa901

Git version = 2.5.0
OSX 10.10.4

cmurphy:
The commit to add a --project-root flag mistakenly tried to use the
opts variable passed in, which is not actually the options hash of
command line arguments. The project_root variable is exposed in the
function, so use that instead.

Closes #55

### DIFF
--- a/lib/modulesync/git.rb
+++ b/lib/modulesync/git.rb
@@ -36,7 +36,7 @@ module ModuleSync
       end
 
       # Repo needs to be cloned in the cwd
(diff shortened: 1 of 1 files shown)


nibalizer:
:+1:

fatmcgav:
Looks good :) :+1:

PR #62, opened 2015-10-23
Ability to have modules with a custom namespace

electrical:
With this change you can have modules inside the same managed_modules
file but with a custom namespace for exceptions.
This can be handy if you have modules across different namespaces but
want to manage them in the same way

### DIFF
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -50,6 +50,14 @@ module ModuleSync
     managed_modules
   end
 
+  def self.module_name(module_name, default_namespace)
+    if module_name.include?('/')
+      namespace, module_name = module_name.split('/')
+    else
+      return [default_namespace, module_name]
+    end
+  end
+
   def self.run(options)
     options = config_defaults.merge(options)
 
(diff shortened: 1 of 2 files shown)


bastelfreak:
thanks for the contribution!

lib/modulesync.rb
```diff
@@ -50,6 +50,14 @@ def self.managed_modules(path, filter)
     managed_modules
   end
 
+  def self.module_name(module_name, default_namespace)
+    if module_name.include?('/')
+      namespace, module_name = module_name.split('/')
```
igalic: this reuse/reassignment/return of the module_name feels… dirty…

lib/modulesync.rb
```diff
@@ -50,6 +50,14 @@ def self.managed_modules(path, filter)
...
+    else
+      return [default_namespace, module_name]
+    end
```
igalic: this whole function feels very unrubylike…

igalic:
```ruby
def self.module_name(module_name, default_namespace)
  return [default_namespace, module_name] unless module_name.include?('/')
  return ns, mod = module_name.split('/')
end
```

PR #65, opened 2015-12-11
Refactoring the Hook module into a class

danzilio:
This commit moves ModuleSync::Hook from a module to a class. It also
adds the ability to define arbitrary arguments to add to the hook. I've
also broken up the `run` method. Instead of calling a monolithic `run`
method we now call `update` or `hook` methods based on the CLI option.

### DIFF
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -55,54 +55,60 @@ module ModuleSync
     ns, mod = module_name.split('/')
   end
 
-  def self.run(options)
+  def self.hook(options)
+    hook = Hook.new(HOOK_FILE, options)
+
+    case options[:hook]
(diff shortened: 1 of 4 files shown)


lib/modulesync.rb
```diff
@@ -55,54 +55,60 @@ def self.module_name(module_name, default_namespace)
     ns, mod = module_name.split('/')
   end
 
-  def self.run(options)
-    options = config_defaults.merge(options)
+  def self.hook(options)
```
igalic: good names are good.

bastelfreak:
discussed it with @jyaworski, we both think that this is fine

igalic:
\o/

PR #66, opened 2015-12-11
Removing constants and clarifying project_root

danzilio:
This PR removes the `Constants` module and moves towards a more composable
configuration model. I've made some changes to the Cucumber features so
the tests fail more descriptively. I've also added a `Config` class that
takes over much of the responsiblity of the `Constants` module. I've moved
responsiblity for maintaining configuration defaults to the `Config` class
and removed them from the CLI class. I've also renamed `project_root`
because it was confusing. Now the project_root refers to the directory
where the `modulesync.yml` file lives. The previous `project_root` referred
to the `modules/` directory and so I have very creatively renamed it to
`modules_dir`.


lib/modulesync/config.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    def defaults
+      {
+        :moduleroot_dir       => File.join(project_root, 'moduleroot/'),
+        :modules_dir          => File.join(project_root, 'modules'),
+        :defaults_file        => File.join(project_root, 'config_defaults.yml'),
+        :managed_modules_file => File.join(project_root, 'managed_modules.yml'),
+        :hook_file            => File.join(project_root, '.git/hooks/pre-push'),
+        :branch               => 'master',
+        :namespace            => 'puppetlabs',
```
bastelfreak: do we want puppetlabs or our own here?

bastelfreak:
:+1:  from my side if the namespace gets discussed. This is much ruby voodoo so I don't want  to merge that alone, could somebody else review it please?

lib/modulesync/config.rb
danzilio: Hmmm...good question...

igalic: keep for now

lib/modulesync/cli.rb
```diff
@@ -1,54 +1,53 @@
...
-      option :bump, :type => :boolean, :desc => 'Bump module version to the next minor', :default => false
-      option :changelog, :type => :boolean, :desc => 'Update CHANGELOG.md if version was bumped', :default => false
-      option :tag, :type => :boolean, :desc => 'Git tag with the current module version', :defalut => false
+      option :amend, :type => :boolean, :desc => 'Amend previous commit'
+      option :force, :type => :boolean, :desc => 'Force push amended commit'
+      option :noop, :type => :boolean, :desc => 'No-op mode'
+      option :offline, :type => :boolean, :desc => 'Do not run any Git commands. Allows the user to manage Git outside of ModuleSync.'
+      option :bump, :type => :boolean, :desc => 'Bump module version to the next minor'
+      option :changelog, :type => :boolean, :desc => 'Update CHANGELOG.md if version was bumped'
+      option :tag, :type => :boolean, :desc => 'Git tag with the current module version'
```
igalic: what happened to these config options' defaults?

danzilio: they've been moved to the defaults method in the config class

igalic: ah. i see. does that mean all boolean options default to false?

danzilio: oh good question. i believe so, or at least they behave falsey so they could be `nil`. i'll double check

bastelfreak:
@danzilio how is the current state here? could you also rebase please?

igalic:
@danzilio reping

danzilio:
@bastelfreak @igalic what do you guys think? should i dust this off?

jyaworski:
I think it's worth a rebase.

jyaworski:
Ping @danzilio

danzilio:
Will try and rebase this week

juniorsysadmin:
@danzilio bump

PR #69, opened 2016-02-10
add a header to any file touched by module sync

tphoney:


mcanevet:
@tphoney what about files types that does not support `#` as commented lines marker?

tphoney:
Any feedback is welcome, not sure if this should be a constant, or optional? 
@mcanevet  thanks for the feedback, i will have a think. There could be a list of files that are exceptions ?

ferventcoder:
@mcanevet that's exactly what I was wondering. :)

tphoney:
@mcanevet @ferventcoder in config_defaults.yml we could have a key of header, which would allow you to disable the header eg:

Gemfile:
  header: false 
    ':development, :unit_tests':
....

ferventcoder:
That kind of defeats the purpose though IMHO. I want to know about everywhere that goes and a header should support the comment based on the place it is going's proper comment. Perhaps a property bag for those files that don't use `#`?

ferventcoder:
Updated my previous comment to clarify what I was saying :)

tphoney:
Yeah, where would you suggest that property bag live, would that be in modulesync as a constant, or in the modulesync_config ? @ferventcoder

ferventcoder:
I think both locations. I think there are some that are known that would be here, but I think it should be able to be overridden in modulesync_config.

ferventcoder:
Just as we have the ability to override things all the way down to the module, maybe even it goes down further if that modulesync_config wants it to.

tphoney:
@ferventcoder do we want the ability to specify the comment symbol '#' or just have file_types that are ignored

ferventcoder:
I think in a header text, you would leave out the comment symbol and then another operation uses that comment symbol when it combines it to the files. And it looks at the property bag and the file type to see if it should change the symbol from a default of `#`. Thoughts?

tphoney:
So a user would be able to specify if a file type should have a different symbol. 

*.cnf   = '+'
*.java  = '='

What file  you expect the property bag to live in ?

ferventcoder:
> What file you expect the property bag to live in ?

No idea. That's up for determination. :)

tphoney:
i will make some changes tomorrow, and push it up. thanks for the feedback

tphoney:
So after a long conversation with @DavidS we decided that solving this programmatically does not make sense. There are too many edge cases, no header / custom header / custom symbol / file types.
The easiest solution is to manually put a header this into https://github.com/puppetlabs/modulesync_configs for each file.
I will create a PR for the rubocop fixes that i uncovered.

PR #72, opened 2016-02-15
Pull git_base into a CLI param

DavidS:
This allows flexible overriding of the git base, e.g. when a CI job needs
to use github's https access, when no SSH credentials are available.


bastelfreak:
the travis fail is based on a broken rubocop conf, all checks passed.

PR #74, opened 2016-02-15
Only calculate a single set of configs using the file's values

DavidS:
In 98aac8cca6a463b33e1d49836bc3315e4eed092a it was overlooked that
per-file defaults and per-file module_configs need to be considered.

This fails the cucumber tests, but I'm running out of daylight today and wanted to get up a discussion base.

### DIFF
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -15,7 +15,7 @@ module ModuleSync
       :project_root         => 'modules/',
       :managed_modules_conf => 'managed_modules.yml',
       :configs              => '.',
-      :tag_pattern          => '%s',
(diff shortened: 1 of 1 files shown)


danzilio:
ahhhhhh i see now.

i wonder if we should just offload our configs to hiera

DavidS:
@danzilio would be great if you could have a look.
the fix is not really pretty, but it does the job.

DavidS:
This is now sucessfully running a preview for our configs in travis: https://travis-ci.org/puppetlabs/modulesync_configs/builds/109834143

danzilio:
Thanks @DavidS!!

PR #77, opened 2016-05-17
Expose config variables to templates

davidmogar:
We needed a way to get config variables like git_base and namespace in template files so we can generate, as an example, better README files.

This change adds those variables to @configs hash. It also add some documentation and a new test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,6 +337,16 @@ current date, bumped (minor) version, and commit message.
 If `CHANGELOG.md` is absent in the repository, nothing will happen.
 
 
+#### Defining templates
+
+As commented, files within moduleroot directory can be flat files or ERB templates. These files have direct access to @configs hash, which gets values from config_defaults.yml file and from the module being processed:
+
+```
+<%= @configs[:git_base] %>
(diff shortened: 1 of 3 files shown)


igalic:
hm… what's up with them tests?

davidmogar:
@igalic Not sure. The new test is working but two of the previous ones were failing. I haven't checked them to be honest.

davidmogar:
Any progress?

davidmogar:
Almost a month has passed. Any info?

danzilio:
hi @davidmogar. sorry for the delay on this. i'm looking at this now and will have an update shortly

davidmogar:
Perfect. Thanks for the feedback ;)

danzilio:
Ok the test failures here are unrelated. They're basically due to some poorly designed tests. The tests that are failing are using puppetlabs/stdlib as a fixture, but the assumptions made about that module in the tests are no longer valid. I'm going to merge this and fix the tests in a separate PR.

davidmogar:
Perfect, thank you.

PR #88, opened 2016-11-23
Add -x to skip specified repositories

DavidS:
This can be useful to avoid snowflakes.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,0 +1,2 @@
+--color
+--format=documentation
--- a/README.md
+++ b/README.md
@@ -265,12 +265,21 @@ pre_commit_script: openstack-commit-msg-hook.sh
 
 #### Filtering Repositories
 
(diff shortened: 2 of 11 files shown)


bmjen:
👍

rnelson0:
I do like this feature! However, would -x or -e be a better flag? -v matches the grep flag but I think more users expect it to mean verbose these days.

alexjfisher:
Other than perhaps the choice of `-v`, LGTM.  The relevant tests pass.  Build is failing due to unrelated rubocop failures.

eg
```
lib/modulesync.rb:93:5: C: Block has too many lines. [40/25]
```
Probably best if these are fixed in a separate PR though.  The `managed_modules.each` block probably *is* too complicated, so I guess it's better not to disable the cop.

bastelfreak:
thanks @DavidS !

PR #91, opened 2016-12-14
(maint) fixes delete issue

eputnam:
Only add files to a commit that exist locally. 

Scenario: A file has already been removed from one or more managed modules and is also added to the modulesync_config as a file to delete. When modulesync is run, an error is thrown saying i've "removed" this file but I can't add it to the commit because it didn't exist to begin with.

### DIFF
--- a/features/update.feature
+++ b/features/update.feature
@@ -170,6 +170,28 @@ Feature: update
       """
     And the exit status should be 0
 
+  Scenario: Setting a non-existent file to deleted
+    Given a file named "managed_modules.yml" with:
+      """
+      ---
+        - puppet-test
+      """
+    And a file named "modulesync.yml" with:
+      """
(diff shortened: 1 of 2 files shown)


bastelfreak:
hey @eputnam, thanks for the PR! Can you take a look at the failling tests and maybe add new one for the change?

eputnam:
yeah totally. i'm not sure it actually works yet anyway.

eputnam:
i'm having trouble reproducing the current travis failures. would someone be able to take a look?

bastelfreak:
I've no clue how this magic works :( @petems or @danzilio do you have any ideas?

bmjen:
I rekicked the job and it's green now.

bastelfreak:
Thanks!

PR #93, opened 2017-01-04
allow .erb extension

james-powis:
This change allows for files to be named with or without .erb extension
This also addresses when there are numerous .rubocop.yml files which
may error if containing erb code.

reference: https://github.com/voxpupuli/modulesync_config/issues/295

### DIFF
--- a/lib/modulesync/renderer.rb
+++ b/lib/modulesync/renderer.rb
@@ -10,8 +10,14 @@ module ModuleSync
     end
 
     def self.build(from_erb_template)
-      erb_obj = ERB.new(File.read(from_erb_template), nil, '-')
-      erb_obj.filename = from_erb_template.chomp('.erb')
+      from_erb_template = from_erb_template.chomp('.erb')
+      template_file = if File.exist?("#{from_erb_template}.erb")
+                        "#{from_erb_template}.erb"
+                      else
+                        from_erb_template
+                      end
+      erb_obj = ERB.new(File.read(template_file), nil, '-')
+      erb_obj.filename = from_erb_template
       erb_obj.def_method(ForgeModuleFile, 'render()')
       erb_obj
     end
@@ -27,7 +33,7 @@ module ModuleSync
     def self.sync(template, to_file)
       path = to_file.rpartition('/').first
       FileUtils.mkdir_p(path) unless path.empty?
-      File.open(to_file, 'w') do |file|
+      File.open(to_file.chomp('.erb'), 'w') do |file|
         file.write(template)
       end
     end
--- a/lib/modulesync/settings.rb
+++ b/lib/modulesync/settings.rb
@@ -12,8 +12,16 @@ module ModuleSync
       @additional_settings = additional_settings
     end
 
+    def lookup_config(hash, filename)
+      hash[filename] || {}
+    end
+
     def build_file_configs(filename)
-      global_defaults.merge(defaults[filename] || {}).merge(module_defaults).merge(module_configs[filename] || {}).merge(additional_settings)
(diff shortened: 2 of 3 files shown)


lib/modulesync/settings.rb
```diff
@@ -12,8 +12,17 @@ def initialize(global_defaults, defaults, module_defaults, module_configs, addit
       @additional_settings = additional_settings
     end
 
+    def lookup_config(hash, filename)
+      fn = filename.chomp('.erb')
+      hash[fn] || hash["#{fn}.erb"] || {}
```
domcleal: `"#{fn}.erb"` is simply `filename`

lib/modulesync/renderer.rb
```diff
@@ -10,8 +10,14 @@ def initialize(configs = {})
     end
 
     def self.build(from_erb_template)
```
domcleal: Change this variable name to `template_file` and then you don't need the `File.exist?` check below to work backwards to what was passed in, it's much simpler. Then `from_erb_template = template_file.chomp('erb')`.

lib/modulesync/settings.rb
james-powis: Assume we are talking about `rubocop.yml.erb` producing `.rubocop.yml`

In my imaginary world of best practices (which conforms to yours in this case) the settings files should key off of `.rubocop.yml.erb` and not `.rubocop.yml` this conditional return was to "just work" when someone keys off of .rubocop.yml (even though they are messing with .erb)... I am happy to implement what you suggested though if you don't see any value in permitting consumer sloppiness.

lib/modulesync/renderer.rb
james-powis: See other discussion in this PR this is just handling consumer sloppiness.

james-powis:
Test have been written, I included the agnostic settings / file name with or without .erb as well.

lib/modulesync/settings.rb
domcleal: The change I suggested doesn't change the logic, I'm not questioning that. I'm only suggesting that the code currently _removes_ the .erb extension and stores it as `fn`, then appends it again on this line. The `"#{fn}.erb"` code can simply use `filename`.

lib/modulesync/renderer.rb
domcleal: Same comment, this is a simplification to the code, not questioning the logic.

domcleal:
The branch is also conflicting with master.

lib/modulesync/renderer.rb
james-powis: using this code and the tests packaged with this PR:
```ruby
    def self.build(template_file)
      erb_obj = ERB.new(File.read(template_file), nil, '-')
      erb_obj.filename = template_file.chomp('.erb')
      erb_obj.def_method(ForgeModuleFile, 'render()')
      erb_obj
    end
```

returns the following failures:
cucumber features/update.feature:628 # Scenario: Using .erb extension template but settings ommit .erb
cucumber features/update.feature:663 # Scenario: Using non-.erb extension template but settings use .erb

Please advise

lib/modulesync/settings.rb
james-powis: I have implemented this ask.

lib/modulesync/renderer.rb
domcleal: Ah sorry, I see what it's trying to do now - opportunistically use a .erb extension template even when called for a filename _without_ the ERB extension. I assumed that `template_file` always was given the ERB extension, but that's not the case.

domcleal:
Merged, thank you @james-powis!

PR #96, opened 2017-01-06
Fix globals symbol

hunner:
Without this commit, the globals entry defaults to a symbol `:global`
and this is initially passed through to `managed?()` which then calls
Pathname. Pathname expects a string so this fails.

With this patch we first check if the key is the default globals key
before passing it through, thus allowing globals to work correctly.


james-powis:
LGTM

Issue #97, opened 2017-01-27
A broken sync on one repository should continue on to the next repositories


traylenator:
Currently when module sync fails on one repository for whatever reason e.g auth error it
stops processing any more repositories.

Propose that modulesync should continue to try the next repository.

domcleal:
Fixed in #98.

PR #98, opened 2017-02-01
Handle rendering exception

davidmogar:
If there was a rendering problem while processing a module, the
whole update process was stopped. Now the process will continue for
the remaining modules. Fix issue #97.

### DIFF
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -98,32 +98,38 @@ module ModuleSync
 
     # managed_modules is either an array or a hash
     managed_modules.each do |puppet_module, module_options|
-      puts "Syncing #{puppet_module}"
-      namespace, module_name = self.module_name(puppet_module, options[:namespace])
-      unless options[:offline]
-        git_base = options[:git_base]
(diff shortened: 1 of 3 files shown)


lib/modulesync.rb
```diff
@@ -98,32 +128,11 @@ def self.update(options)
...
-
-      if options[:noop]
-        Git.update_noop(module_name, options)
-      elsif !options[:offline]
-        Git.update(module_name, files_to_manage, options)
+      begin
+        manage_module(puppet_module, module_files, module_options, defaults, options)
+      rescue
+        raise unless options[:skip_broken]
+        puts "Skipping #{puppet_module} as update process failed"
```
domcleal: Could you add some information about the exception here, so it's easier to diagnose the issue? At least `e.message`, but also possibly `e.backtrace.join("\n")`.

davidmogar: Added a similar error message to the one showed while rendering files. Think is enough in this case. I also added the test for skip_broken feature. Let me know if you want any other change.

domcleal:
Merged, thanks @davidmogar.

PR #100, opened 2017-02-08
Fix invalid check

### LINKED ISSUE #99: Files not been added correctly before pushing
As I [commented in the commit](https://github.com/voxpupuli/modulesync/commit/e6b433262e16ac8894806b5adc3532fe032bd681#commitcomment-20783215) itself, I think last change to git.py file added a bug.

[This loop](https://github.com/voxpupuli/modulesync/blob/e6b433262e16ac8894806b5adc3532fe032bd681/lib/modulesync/git.rb#L107) is checking for files in the current directory and not in the repository.

davidmogar:
The code was checking for repository files in the wrong path (current
directory). Fix #99.

### DIFF
--- a/lib/modulesync/git.rb
+++ b/lib/modulesync/git.rb
@@ -104,7 +104,7 @@ module ModuleSync
       files.each do |file|
         if repo.status.deleted.include?(file)
           repo.remove(file)
-        elsif File.exist?(file)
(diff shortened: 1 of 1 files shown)


domcleal:
Merged, thanks @davidmogar.

domcleal:
I've added a test to demonstrate the bug in #101 if you wouldn't mind reviewing it.

PR #101, opened 2017-02-08
Add test to check updated files are committed and pushed

domcleal:
Refs #99

### DIFF
--- a/features/update.feature
+++ b/features/update.feature
@@ -126,6 +126,44 @@ Feature: update
       source 'https://somehost.com'
       """
 
+  Scenario: Modifying an existing file and committing the change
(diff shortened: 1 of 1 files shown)


domcleal:
Test is now passing, the git user.name/email settings needed setting in a mocked home directory under Travis CI.

domcleal:
> I think this is not a valid test as Gemfile file already exists in modulesync repository.

For cucumber tests, they're running with a pwd of `tmp/aruba/` and don't use a Gemfile so the file isn't there twice. If you revert the fix then you'll see the test fails.

davidmogar:
My fault. You are right. I was checking the wrong branch while testing.

PR #103, opened 2017-02-09
Add diff before pushing

### LINKED ISSUE #102: Push fails when not using --force
During updates, if `--force` options is not used the push can fail even if there is no changes between branches. More on, it makes no sense to perform that push in the first place.

In our use case, we push to a remote branch and create a merge request. If this request is not accepted before running modulesync again, `msync update` will try to push and the process will fail. Compare the local branch with our remote branch would solve the problem.

davidmogar:
Fix #102


domcleal:
Merged, thanks @davidmogar.

Issue #82, opened 2016-07-14
Specify ruby requirements in gemspec


rnelson0:
Per https://github.com/voxpupuli/modulesync/blob/master/.travis.yml#L7 it looks like this requires ruby 1.9.3. Whatever the required version is, explicitly state this in the gemspec, e.g. `s.required_ruby_version = '>= 1.9.3'`

bastelfreak:
Hi @rnelson0, we just did https://github.com/voxpupuli/modulesync/pull/109 which fixes this. let me know if you need a release with this fix.

rnelson0:
I think it can wait till release 0.8.0.

PR #109, opened 2017-02-13
run on new travis env + drop legacy ruby versions

bastelfreak:


domcleal:
Do you want to also increase/specify the minimum Ruby version to match what's tested (in the gemspec and then in .rubocop.yml)? This would also fix #82.

roidelapluie:
Can we please keep ruby 1.9?

.rubocop.yml
```diff
@@ -1,13 +1,12 @@
 inherit_from: .rubocop_todo.yml
 
 AllCops:
-  TargetRubyVersion: 1.9
+  TargetRubyVersion: 2.1
```
roidelapluie: Please keep 2.0

bastelfreak:
I would really love to drop all the EOL versions like ruby2.0 and ruby2.1, I've a bad feeling about supporting them. However, little progress is better than nothing.

PR #110, opened 2017-02-14
Change git required version

davidmogar:
The new diff functionality in git.rb requires a newer rubygem-git version.

### DIFF
--- a/modulesync.gemspec
+++ b/modulesync.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rspec'
   spec.add_development_dependency 'rubocop'
 
-  spec.add_runtime_dependency 'git', '~>1.2'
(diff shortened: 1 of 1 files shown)


domcleal:
The diff functionality appears to have been there for a long time, and it's been used in other parts of modulesync before. Can you point to which `git` (ruby-git) release contains it? 1.3.0 only seems to drop Ruby 1.8.7 support.

The test seems to run OK on version 1.2.2.

davidmogar:
We tried to use `diff` from `ruby-git 1.2.8` on a machine with `ruby 2.0.0p648`. The function throw an error like this one:
```
git diff '-p' 'qa..origin/modulesync' ''  2>&1:fatal: ambiguous argument '': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like
```
We solved it updating to `ruby-git 1.3.0`.

domcleal:
Merged, thanks @davidmogar.

Issue #45, opened 2015-05-08
consider using libgit2 to avoid issues with wonky clients or configurations


igalic:
https://github.com/schacon/ruby-git/issues/231 <- 2 hours of debugging with 2 two engineers.

rski:
Does this still need to be done? If yes, will it be done with [rugged](https://github.com/libgit2/rugged)?
I could look into producing a prototype over the next few weeks.

igalic:
it's on the roadmap
and any work on it would be very welcome

rski:
Turns out there is one very important thing that ruby-git provides: Authentication with the git client. 
`libgit2` does authentication programmatically, so a user interface (file? command line input?) would need to be implemented to get authentication. 
The problem is that implementing anything for this would just be redoing work that already exists in the actual git client, which is why `ruby-git` just wraps `git`.
I think trying to purge ruby-git's `options` hash before running any git commands might solve the second part of this issue, i.e. wonky configurations. I'm not sure how sane and doable that would be though.

igalic:
@rski thank you very much for the research!

PR #111, opened 2017-02-22
Monkey Patch git ruby gem

glennsarti:
Previously modulesync would throw an error when a file had a line ending warning
emitted during a diff.  This is common on cross platform modules.  This commit
monkey patches the git ruby gem to handle extra text before the actual diff
contents.  A monkey patch was used as the git ruby gem does not seem to be
actively maintained, and it will some time before this issue is resolved and a
new gem available.

Git ruby gem issue is https://github.com/schacon/ruby-git/issues/326

### DIFF
--- a/lib/monkey_patches.rb
+++ b/lib/monkey_patches.rb
@@ -1,0 +1,40 @@
+module Git
+  class Diff
+    # Monkey patch process_full_diff until https://github.com/schacon/ruby-git/issues/326 is resolved
+    def process_full_diff
+      defaults = {
+        :mode => '',
+        :src => '',
+        :dst => '',
+        :type => 'modified'
+      }
+      final = {}
+      current_file = nil
+      full_diff_utf8_encoded = @full_diff.encode("UTF-8", "binary", {
+    :invalid => :replace,
+    :undef => :replace
+  })
+      full_diff_utf8_encoded.split("\n").each do |line|
+        if m = /^diff --git a\/(.*?) b\/(.*?)/.match(line)
+          current_file = m[1]
+          final[current_file] = defaults.merge({:patch => line, :path => current_file})
+        elsif !current_file.nil?
+          if m = /^index (.......)\.\.(.......)( ......)*/.match(line)
+            final[current_file][:src] = m[1]
+            final[current_file][:dst] = m[2]
+            final[current_file][:mode] = m[3].strip if m[3]
+          end
+          if m = /^([[:alpha:]]*?) file mode (......)/.match(line)
+            final[current_file][:type] = m[1]
+            final[current_file][:mode] = m[2]
+          end
+          if m = /^Binary files /.match(line)
+            final[current_file][:binary] = true
+          end
+          final[current_file][:patch] << "\n" + line
+        end
+      end
+      final.map { |e| [e[0], DiffFile.new(@base, e[1])] }
+    end
+  end
+end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,6 +6,7 @@ AllCops:
     - 'vendor/**/*'
     - 'tmp/**/*'
     - 'pkg/**/*'
+    - 'lib/monkey_patches.rb'
 
 Style/HashSyntax:
   Enabled: false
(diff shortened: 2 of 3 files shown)


glennsarti:
Repro case:
* Download a repository which contains a file with CRLF endings on an operating system that uses LF endings
* Extracting the diff will throw a `NoMethodError: undefined method '[]' for nil:NilClass` error in method `process_full_diff`

Example output;
```
Overriding any local changes to repositories in modules
Not managing Rakefile in puppetlabs-chocolatey
Not managing spec/spec_helper.rb in puppetlabs-chocolatey
Using no-op. Files in puppetlabs-chocolatey may be changed but will not be committed.
Files changed:
Error while updating puppetlabs-chocolatey
bundler: failed to load command: msync (/mnt/c/Source/modulesync_configs/.bundle/bash/ruby/2.3.0/bin/msync)
NoMethodError: undefined method `[]' for nil:NilClass
  /mnt/c/Source/modulesync_configs/.bundle/bash/ruby/2.3.0/gems/git-1.3.0/lib/git/diff.rb:151:in `block in process_full_diff'
  /mnt/c/Source/modulesync_configs/.bundle/bash/ruby/2.3.0/gems/git-1.3.0/lib/git/diff.rb:134:in `each'
  /mnt/c/Source/modulesync_configs/.bundle/bash/ruby/2.3.0/gems/git-1.3.0/lib/git/diff.rb:134:in `process_full_diff'
  /mnt/c/Source/modulesync_configs/.bundle/bash/ruby/2.3.0/gems/git-1.3.0/lib/git/diff.rb:109:in `process_full'
  /mnt/c/Source/modulesync_configs/.bundle/bash/ruby/2.3.0/gems/git-1.3.0/lib/git/diff.rb:68:in `each'
  /mnt/c/Source/modulesync_configs/.bundle/bash/ruby/2.3.0/bundler/gems/modulesync-899483f48304/lib/modulesync/git.rb:162:in `update_noop'
  /mnt/c/Source/modulesync_configs/.bundle/bash/ruby/2.3.0/bundler/gems/modulesync-899483f48304/lib/modulesync.rb:114:in `manage_module'
  /mnt/c/Source/modulesync_configs/.bundle/bash/ruby/2.3.0/bundler/gems/modulesync-899483f48304/lib/modulesync.rb:133:in `block in update'
```

The reason it fails because the code at; https://github.com/schacon/ruby-git/blob/master/lib/git/diff.rb#L139-L151 expects the first line to contain the filename whereas raw diff comes back as;

```
warning: CRLF will be replaced by LF in examples/init.pp.
The file will have its original line endings in your working directory.
diff --git a/.gitattributes b/.gitattributes
index 900ea0c..02d4646 100644
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
 #This file is generated by ModuleSync, do not edit.
```

lib/monkey_patches.rb
```diff
@@ -0,0 +1,40 @@
...
+        :src => '',
+        :dst => '',
+        :type => 'modified'
+      }
+      final = {}
+      current_file = nil
+      full_diff_utf8_encoded = @full_diff.encode("UTF-8", "binary", {
+    :invalid => :replace,
+    :undef => :replace
+  })
```
rnelson0: Some funky alignment here.

glennsarti: Yeah...That was a copy-n-paste from the git ruby gem.  It has a combo of spaces and tabs :-(

glennsarti:
@rnelson0 Did you mean an issue on this repo or the git ruby gem?  Already raised schacon/ruby-git#326 to track it.

glennsarti:
And I'm not a fan of monkey patching either, but I couldn't see any other way of getting this done without a major overhaul.  This may also fix https://github.com/voxpupuli/modulesync/issues/45 too

rnelson0:
An issue on this gem, so we don't forget about it.

lib/monkey_patches.rb
rnelson0: If that's from the original it makes sense to keep it.

glennsarti:
Raised https://github.com/voxpupuli/modulesync/issues/112

glennsarti:
Thanks @rnelson0 !

PR #114, opened 2017-04-21
ERB suffix handling

DavidS:


DavidS:
cc @james-powis @puppetlabs/modules @puppetlabs/sdk

james-powis:
All in all everything looks pretty concise and certainly a good cleanup of the code base. Per your description "prefer .erb in filename" certainly looks like it is required, none of the tests (or the code as I am interpreting it) allow for backward compatibility for non-.erb suffixed in module root. Was that the intention?

DavidS:
right, I went back on that decision last minute, and never added a test for
that. I probably also need a negative test for the case this does not
support (i.e. .erb suffix in config keys)

james-powis:
I am really concerned about the level of work required by the community and members outside the community who leverage modulesync who are going to have to refactor because of this... it does not add much complexity to support non-.erb suffixed files, is there any way we can keep that logic but insist on the absence of .erb in the associated config files?

DavidS:
https://github.com/voxpupuli/modulesync/pull/114/files#diff-94f12d4c397c9e0adbe4eb57f1c8dcbeR13 still allows non-erb suffixed templates.

Fixing a template is as easy as `find moduleroot/ -type f -exec git mv {} {}.erb \; && git commit -m "Update all files to .erb"`  I can add that to the CHANGELOG.

DavidS:
@domcleal I've added the additional test, and I've created #115 as a separate PR (re-using the same commits) to make this PR here smaller, should this one need more work.

domcleal:
Merged, thanks @DavidS.

bittner:
The [README still says](https://github.com/voxpupuli/modulesync#defining-templates):

> As commented, files within moduleroot directory can be flat files or ERB templates.

This is now incorrect, right? It should say:

> As commented, files within the moduleroot directory must be ERB templates.

PR?

PR #115, opened 2017-04-28
Rubocop fixes

DavidS:
break out the rubocop fixes from #114 into separate PR to make the diff there smaller.


domcleal:
Thanks @DavidS, merged.

PR #117, opened 2017-05-05
Allow the CLI to override options set in the config file

DavidS:
This follows the generally expected precedence of the CLI options
overriding values from the config file. In particular, I found this when
trying to create a travis job for modulesync_config
(see https://travis-ci.org/voxpupuli/modulesync_config/jobs/229043830).

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,7 +234,8 @@ Available parameters for modulesync.yml
 * namespace : Namespace of the projects to manage (Default: 'puppetlabs')
 * branch : Branch to push to (Default: 'master')
 * remote_branch : Remote branch to push to (Default: Same value as branch)
-* pre_commit_script : A script to be run before commiting (ie. contrib/myfooscript.sh)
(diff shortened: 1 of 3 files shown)


domcleal:
The `remote_branch` command line option should default to the config file value too, according to README.md. VP's modulesync_config also has `message` set, but this isn't documented - it might be reasonable to default the `message` option to the config file value if set and document it.

Lastly, `pre_commit_script` is a documented config file option but it isn't exposed through the CLI arguments. I think as a result of removing the `config.merge!(config_file)`, this will no longer be populated from the config file. It'll either need to be loaded into the options hash from `defaults` or exposed as a CLI option, defaulting to the config file value.

It might be worth checking through the source for any more undocumented options - sorry!

DavidS:
@domcleal fixed and travissed!

domcleal:
Thanks @DavidS!

PR #119, opened 2017-06-01
(maint) Preserve original line endings in a file

glennsarti:
Previously when rendering a file the line endings would be converted by ruby
to the OS specific style e.g. CRLF on Windows.  However as many modules are
cross platform it is important to preserve the line ending style from the
original moduleroot file e.g. a .BAT file is for windows and should always use
CRLF, whether it's on a Linux of Windows system.  This commit changes the file
opening mode to use binary (which stops the line ending conversion) and forces
all file encoding to UTF-8 which is the default already.

### DIFF
--- a/lib/modulesync/renderer.rb
+++ b/lib/modulesync/renderer.rb
@@ -33,7 +33,7 @@ module ModuleSync
     def self.sync(template, target_name)
       path = target_name.rpartition('/').first
       FileUtils.mkdir_p(path) unless path.empty?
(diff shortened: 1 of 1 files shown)


domcleal:
This looks reasonable, would you be able to add a test to `features/` to show it works? (Probably by using an ERB template with CRLF line endings, and showing they're preserved.)

glennsarti:
Apologies.  Forgot about this PR.  Will add tests

PR #123, opened 2017-06-26
Fix link to config repo in README

zabacad:
Without this change, the name and link to Voxpupuli's config repo is incorrect.


bastelfreak:
Thanks!

PR #130, opened 2017-09-04
Using repo.branch.name always returns master, use repo.current_branch…

### LINKED ISSUE #129: ModuleSync::Git.switch_branch can never switch to master
`repo.branch` is meant to return a `Git::Branch` object for whichever branch name is passed to it.  It defaults to `master` so this guard clause returns when the branch parameter is also `master`: https://github.com/voxpupuli/modulesync/blob/master/lib/modulesync/git.rb#L22

```
irb(main):001:0> require 'modulesync'
No config file under modulesync.yml found, using default values
=> true
irb(main):002:0> g = Git.open('.')
=> #<Git::Base:0x005fab0f6c24c8 @logger=nil, @working_directory=#<Git::WorkingDirectory:0x005fab0f6bfc78 @path="/home/sean/dev/modulesync">, @repository=#<Git::Repository:0x005fab0f6bfb38 @path="/home/sean/dev/modulesync/.git">, @index=#<Git::Index:0x005fab0f6bf840 @path="/home/sean/dev/modulesync/.git/index">>
irb(main):003:0> g.branch.name
=> "master"
irb(main):004:0> g.current_branch
=> "use_git_current_branch"
irb(main):005:0> ModuleSync::Git.switch_branch(g, 'master')
=> nil
irb(main):006:0> g.current_branch
=> "use_git_current_branch"
```

I think line 22 should use `repo.current_branch` to compare against the passed in branch name.  After changing the code and reloading the file I get the expected behavior:
```
irb(main):007:0> load 'lib/modulesync/git.rb'
=> true
irb(main):008:0> ModuleSync::Git.switch_branch(g, 'master')
Switching to branch master
=> "Switched to branch 'master'\nM\tlib/modulesync/git.rb\nYour branch is up-to-date with 'origin/master'."
irb(main):009:0> g.current_branch
=> "master"
```

I'm willing to file a PR to fix this.

asedge:
… for the guard clause instead so ModuleSync::Git.switch_branch can actually switch to the master branch when it is not the current branch.

Fixes #129

### DIFF
--- a/lib/modulesync/git.rb
+++ b/lib/modulesync/git.rb
@@ -19,7 +19,7 @@ module ModuleSync
     end
 
     def self.switch_branch(repo, branch)
-      return if repo.branch.name == branch
+      return if repo.current_branch == branch
 
       if local_branch_exists?(repo, branch)
         puts "Switching to branch #{branch}"


rnelson0:
Thanks for finding and fixing this!

PR #135, opened 2017-11-04
Make rubocop happy

ekohl:


lib/modulesync.rb
```diff
@@ -81,7 +81,7 @@ def self.manage_file(filename, settings, options)
         erb = Renderer.build(templatename)
         template = Renderer.render(erb, configs)
         Renderer.sync(template, module_file(options[:project_root], module_name, filename))
-      rescue
+      rescue # rubocop:disable Lint/RescueWithoutErrorClass
```
alexjfisher: Isn’t the fix easier than disabling the cop?

ekohl: In this particular case (catch, log, reraise) I think the cop is a false positive.

alexjfisher: So `rescue StandardError` would be wrong as the specific sub class would no longer be raised?

ekohl: I don't know Ruby that well but the intention is to catch everything.

alexjfisher: So I think `rescue` and `rescue StandardError` are 100% equivalent and if the intention really is to catch (almost [1]) everything (which I think it is), rubocop wants you to be explicit about this.  There's some recent discussion I found [here](https://github.com/bbatsov/rubocop/issues/4851)
The cop might not be very useful, but is probably doing exactly what it was designed to (not a false positive).

[1] - `rescue Exception` would catch *everything*.

Issue #122, opened 2017-06-07
Files without ERB extension not syncing


cdenneen:
Leaving .rb extension
```
 ~/src/gitlab/puppet/modulesync_configs  master ?  msync update -m "Test change" --noop
Syncing puppet-pes3backup
Cloning repository fresh
Cloning from https://gitlab.domain.local/puppet/puppet-pes3backup.git
No config file under modules/puppet-pes3backup/.sync.yml found, using default values
Using no-op. Files in puppet-pes3backup may be changed but will not be committed.
Files changed:
Files added:


--------------------------------
Syncing puppet-monitoring_scripts
Cloning repository fresh
Cloning from https://gitlab.domain.local/puppet/puppet-monitoring_scripts.git
No config file under modules/puppet-monitoring_scripts/.sync.yml found, using default values
Using no-op. Files in puppet-monitoring_scripts may be changed but will not be committed.
Files changed:
Files added:


--------------------------------
Syncing puppet-profile_base
Cloning repository fresh
Cloning from https://gitlab.domain.local/puppet/puppet-profile_base.git
No config file under modules/puppet-profile_base/.sync.yml found, using default values
Using no-op. Files in puppet-profile_base may be changed but will not be committed.
Files changed:
Files added:


--------------------------------
 ~/src/gitlab/puppet/modulesync_configs  master ?  ls -la modules/*/spec/spec_helper_local.rb
-rw-r--r--  1 cdenneen  staff   340B Jun  6 15:09 modules/puppet-monitoring_scripts/spec/spec_helper_local.rb
-rw-r--r--  1 cdenneen  staff    54B Jun  6 15:09 modules/puppet-profile_base/spec/spec_helper_local.rb
```
Changing extension to `erb`
```
 ~/src/gitlab/puppet/modulesync_configs  master ?  cp moduleroot/spec/spec_helper_local.rb moduleroot/spec/spec_helper_local.rb.erb
 ~/src/gitlab/puppet/modulesync_configs  master ?  msync update -m "Test change" --noop
Syncing puppet-pes3backup
Overriding any local changes to repositories in modules
No config file under modules/puppet-pes3backup/.sync.yml found, using default values
Using no-op. Files in puppet-pes3backup may be changed but will not be committed.
Files changed:
Files added:
spec/spec_helper_local.rb


--------------------------------
Syncing puppet-monitoring_scripts
Overriding any local changes to repositories in modules
[...]

PR #136, opened 2017-12-01
issues-122 Sync files that do not conform to *.erb

tphoney:


tphoney:
This will resolve #122

tphoney:
@DavidS can you have a look

ekohl:
If you support both, would it make sense to only consider files with.```.erb``` extensions templates?

DavidS:
@ekohl that was the goal of my original change. Between push-back from others against that, and this bug here, we're not there yet.

tphoney:
@ekohl @DavidS does this go any further to address your concerns.

ekohl:
I wonder if this should directly implement the non-templated part. I'm afraid the partial implementation might cause confusion.

alexjfisher:
@ghoneycutt I’ve noticed you’ve approved this PR, but since it’s been a year, I’d like to get feedback from @ekohl and @tphoney before merging.

PR #137, opened 2018-01-03
Remove hard coding of managed_modules.yml which means that options pa…

asedge:
…ssed to ModuleSync.update can override the filename.  Also improve parameter/variable name and error message in ModuleSync.managed_modules.

**Explanation of why I need/want this:**
I have a use case where I need to do multiple ModuleSync runs concurrently and they may end up running in the same directory on the same machine.  I'm able to override almost everything to make this possible except `:managed_modules_conf` is useless because `managed_modules.yml` is currently hardcoded in `ModuleSync.update`.

### DIFF
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -43,10 +43,10 @@ module ModuleSync
     local_files.map { |file| file.sub(/#{path}/, '') }
   end
 
-  def self.managed_modules(path, filter, negative_filter)
-    managed_modules = Util.parse_config(path)
+  def self.managed_modules(config_file, filter, negative_filter)
+    managed_modules = Util.parse_config(config_file)
     if managed_modules.empty?
-      puts "No modules found at #{path}. Check that you specified the right configs directory containing managed_modules.yml."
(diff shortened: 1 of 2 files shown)


asedge:
@ekohl I will.  I haven't written any cucumber tests, would spec tests suffice?

ekohl:
I think a spec tests are good enough for now. We can always add cucumber tests if it breaks.

asedge:
@ekohl I added a spec test.  Have a look and let me know if you think I need to do more.  Thanks!

ekohl:
It looks good. Travis is blowing up because we update the gems which then breaks. I've submitted https://github.com/voxpupuli/modulesync/pull/138 which I think should fix it.

ekohl:
With #138 merged, could you rebase?

asedge:
Done.

asedge:
@ekohl Waiting on anything else before merge?  Thanks!

ekohl:
@asedge usually I give it a bit of time between approving and merging when I'm somewhat new to a project to give other people a chance to jump in with their review. Sufficient time has passed now though :)

asedge:
I can appreciate that, thanks!

PR #138, opened 2018-01-04
Do not update gems before_install on Travis

ekohl:
This breaks bundler on various Ruby versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,6 @@ sudo: false
 language: ruby
 cache: bundler
 dist: trusty
-before_install:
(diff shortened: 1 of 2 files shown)


bastelfreak:
Shouldn't we just drop ruby 2.0? It is EOL since such a long time.

ekohl:
@bastelfreak if it still works, I'm fine with keeping it in now. This build only failed because the ruby 2.4 but had connection timeouts to various sources (apt repos, rubygems) so I restarted it.

ekohl:
💚

PR #139, opened 2018-01-06
Closes #132. Add new option to produce a failure exit code on warning…

### LINKED ISSUE #132: Can skip_broken still return a non-zero exit?
I would like to complete the entire run but have ModuleSync still report there was an error so I can take some action when it finishes. Could we find a way to support this behavior? Willing to file a PR.

asedge:
…s.  Only useful when using --skip_broken option to process all modules but needing feedback that problems were encountered.

### DIFF
--- a/features/update.feature
+++ b/features/update.feature
@@ -258,6 +258,34 @@ Feature: update
     When I run `msync update --noop -s`
     Then the exit status should be 0
 
+  Scenario: Using skip_broken and fail_on_warnings options with invalid files
+    Given a file named "managed_modules.yml" with:
+      """
+      ---
+        - puppet-test
+      """
+    And a file named "modulesync.yml" with:
+      """
+      ---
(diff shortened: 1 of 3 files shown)


asedge:
Odd, there are a bunch of failures but only for the Ruby 2.3.3 build.  Whoa and looking at the build time it took an astonishing 21 minutes to run where as the other versions of Ruby only took around 2 minutes.  Do you have the ability to re-run that build?

rnelson0:
I restarted that job but I'm not sure the failures are related to the duration. We shall see.

ekohl:
I saw other Travis builds fail due to timeouts as well totally unrelated to the actual code under test. Looks like it passed now.

rnelson0:
Weird. Well you learn something new every day :D

ekohl:
Thanks @asedge!

PR #141, opened 2018-01-07
Fixes #140. Reraise exception rather than exit so we can rescue a der…

### LINKED ISSUE #140: Skip broken option does not work when git push fails
[This commit](https://github.com/voxpupuli/modulesync/commit/78a0b42a992448c68f9df03871509ad558a0011e) added `exit(1)` which raises `SystemExit`.  [This rescue](https://github.com/voxpupuli/modulesync/blob/master/lib/modulesync.rb#L136) will rescue a `StandardError` but not a `SystemExit`.  A few options on how to fix this.  Simplest might be to just explicitly  `rescue StandardError, SystemExit`.  Alternatively, instead of `exit(1)` we could raise a custom exception derived from `StandardError`.

Thoughts?
  
Edit: Just looked at the diff again from the original commit and realized that this has been a problem for longer than that commit.

asedge:
…ived StandardError when using skip_broken option.  Also, must quote `git_base` option in feature setup, otherwise the YAML parser mistakenly interprets `git@github.com:` as a hash key (I think).

### DIFF
--- a/features/update.feature
+++ b/features/update.feature
@@ -34,6 +34,32 @@ Feature: update
     Given I run `cat modules/puppet-test/test`
     Then the output should contain "aruba"
 
+  Scenario: Using skip_broken option and adding a new file to repo without write access
+    Given a file named "managed_modules.yml" with:
+      """
+      ---
+        - puppet-test
+      """
+    And a file named "modulesync.yml" with:
+      """
+      ---
+        namespace: maestrodev
+        git_base: 'git@github.com:'
+      """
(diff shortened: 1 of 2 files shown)


asedge:
@ekohl I suppose I figured out how to write Cucumber tests... :smile:

asedge:
@bastelfreak Anything else needed here to get this merged?

ekohl:
Thanks!

asedge:
@ekohl Thanks!

PR #144, opened 2018-02-02
Monkey patch ls_files

### LINKED ISSUE #143: fatal: empty string is not a valid pathspec
This is more of an informational issue, so feel free to close as appropriate but when running git 2.16.1, you'll now get the following error:

```
Git::GitExecuteError: git '--git-dir=/home/user/modulesync-config/modules/puppet-mymodule/.git' '--work-tree=/home/user/modulesync-config/modules/puppet-mymodule' ls-files '--stage' ''  2>&1:fatal: empty string is not a valid pathspec. please use . instead if you meant to match all paths
```

I rolled back to 2.15.1 and the error goes away.

This relates to https://github.com/ruby-git/ruby-git/pull/320 in the git ruby gem.

davewongillies:
Monkey patch ls_files until ruby-git/ruby-git#320 is resolved.

Fixes #143

### DIFF
--- a/lib/monkey_patches.rb
+++ b/lib/monkey_patches.rb
@@ -37,4 +37,19 @@ module Git
       final.map { |e| [e[0], DiffFile.new(@base, e[1])] }
     end
   end
+
+  class Lib
+    # Monkey patch ls_files until https://github.com/ruby-git/ruby-git/pull/320 is resolved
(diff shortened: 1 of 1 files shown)


bastelfreak:
Thanks for this! Had the same issues on my machine...

PR #147, opened 2018-04-16
Add support to submit PRs to GitHub when changes are pushed

joestump:
This adds two new CLI options for the `update` command:

* `--pr` is a boolean value that when set will submit a PR to GitHub if updates have been pushed.
* `--pr-title` is a string that will be used for the title of the PR when submitting a PR to GitHub. `-m` is used as the PR's body. You can set this in `modulesync.yml` using the `pr_title` attribute.
* `--pr-labels` is a string of labels to assign the PR. `pr_labels` in `modulesync.yml` is expected to be a list, but a string should work as well.

A few notes about behavior:

* Will not run if `--noop` is passed.
* Will not push a PR if no files have been changed/pushed. `Git.update` was changed to facilitate this behavior. 
* Users must set a GitHub OAuth token in `GITHUB_TOKEN` for this to work.

### DIFF
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -139,7 +139,7 @@ module ModuleSync
 
         # We only assign labels to the PR if we've discovered a list > 1. The labels MUST
         # already exist. We DO NOT create missing labels.
-        if !pr_labels.empty?
+        unless pr_labels.empty?
           puts "Attaching the following labels to PR #{pr['number']}: #{pr_labels.join(', ')}"
           github.add_labels_to_an_issue(repo_path, pr['number'], pr_labels)
         end
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -96,6 +96,11 @@ module ModuleSync
   end
 
   def self.manage_module(puppet_module, module_files, module_options, defaults, options)
+    if option[:pr] && !GITHUB_TOKEN
+        STDERR.puts 'Environment variable GITHUB_TOKEN must be set to use --pr!'
+        raise unless options[:skip_broken]
+    end
+
     puts "Syncing #{puppet_module}"
     namespace, module_name = module_name(puppet_module, options[:namespace])
     unless options[:offline]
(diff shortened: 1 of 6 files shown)


joestump:
@rnelson0 @bastelfreak I aim to please. There are now `--pr-labels` and `pr_labels` options for attaching labels. Only caveat is that the labels must exist ahead of time and YMMV with odd characters in the labels.

lib/modulesync.rb
```diff
@@ -114,7 +121,31 @@ def self.manage_module(puppet_module, module_files, module_options, defaults, op
...
+      # Git.update() returns a boolean: true if files were pushed, false if not.
+      pushed = Git.update(module_name, files_to_manage, options)
+      return nil unless pushed
+
+      if options[:pr] && GITHUB_TOKEN
+        # We only do GitHub PR work if the GITHUB_TOKEN variable is set in the environment.
+        repo_path = "#{namespace}/#{module_name}"
+        puts "Submitting PR '#{options[:pr_title]}' on GitHub to #{repo_path} - merges #{options[:branch]} into master"
+        github = Octokit::Client.new(:access_token => GITHUB_TOKEN)
+        pr = github.create_pull_request(repo_path, 'master', options[:branch], options[:pr_title], options[:message])
```
joestump: @bastelfreak @rnelson0 should this be `:remote_branch`? I'm thinking yes?

lib/modulesync.rb
```diff
@@ -114,7 +121,31 @@ def self.manage_module(puppet_module, module_files, module_options, defaults, op
     if options[:noop]
       Git.update_noop(module_name, options)
     elsif !options[:offline]
-      Git.update(module_name, files_to_manage, options)
+      # Git.update() returns a boolean: true if files were pushed, false if not.
+      pushed = Git.update(module_name, files_to_manage, options)
+      return nil unless pushed
+
+      if options[:pr] && GITHUB_TOKEN
```
ekohl: I'd say if the `--pr` option is given and no token is specified it should be a hard error, probably up front.

lib/modulesync.rb
```diff
@@ -114,7 +121,31 @@ def self.manage_module(puppet_module, module_files, module_options, defaults, op
...
+        # We only do GitHub PR work if the GITHUB_TOKEN variable is set in the environment.
+        repo_path = "#{namespace}/#{module_name}"
+        puts "Submitting PR '#{options[:pr_title]}' on GitHub to #{repo_path} - merges #{options[:branch]} into master"
+        github = Octokit::Client.new(:access_token => GITHUB_TOKEN)
```
ekohl: Would it be wise to cache this object somewhere? Would that allow reusing the network connection or doesn't the gem use that anyway.

joestump: It looks like the underlying library, Faraday, supports persistent HTTP connections, but it's not entirely clear how to make Octokit use it. This feels like a premature optimization given we're already living with Git connections being torn down each module / shell-out operations running in serial. In my testing I've ran this against low dozens of our repos in a single run and it doesn't feel slower than the existing runs.

joestump:
Sadly it looks like the latest changes broke tests in ways I wouldn't have expected.The code works when I actually run it, but tests are failing.

I'm officially out of my depths WRT Ruby so my vote is to either roll back these changes or someone who knows the tests better than I pick up this PR to complete.

lib/modulesync.rb
ekohl: That's fair. Thanks for considering and let's keep this simple for now.

lib/modulesync.rb
```diff
@@ -120,31 +125,27 @@ def self.manage_module(puppet_module, module_files, module_options, defaults, op
 
     if options[:noop]
       Git.update_noop(module_name, options)
-    elsif !options[:offline]
+    elsif !options[:offline] && options[:pr]
```
rnelson0: I believe the test failure is here. The `pushed` segment on 125/130 was previously run when `!offline`, now it's skipped when `!offline`, unless `pr` is enabled.

PR #148, opened 2018-04-26
Copy file permissions from template to target

arjenz:
I have a couple of executable scripts to sync (like a `config_version.rb`), this patch has msync make the rendered output files have the same permissions as the source .erb. 

`File.chmod` the rendered template to mimic the source .erb file,
to allow (for instance) executable scripts to be synced by modulesync

### DIFF
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -84,10 +84,11 @@ module ModuleSync
       Renderer.remove(module_file(options[:project_root], module_name, filename))
     else
       templatename = local_file(options[:configs], filename)
+      mode = File.stat("#{templatename}.erb").mode
       begin
         erb = Renderer.build(templatename)
         template = Renderer.render(erb, configs)
-        Renderer.sync(template, module_file(options[:project_root], module_name, filename))
+        Renderer.sync(template, module_file(options[:project_root], module_name, filename), mode)
       rescue # rubocop:disable Lint/RescueWithoutErrorClass
         STDERR.puts "Error while rendering #{filename}"
         raise
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -84,7 +84,12 @@ module ModuleSync
       Renderer.remove(module_file(options[:project_root], module_name, filename))
     else
       templatename = local_file(options[:configs], filename)
-      mode = File.stat("#{templatename}.erb").mode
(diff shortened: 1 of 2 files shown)


lib/modulesync/renderer.rb
```diff
@@ -30,12 +30,13 @@ def self.render(_template, configs = {})
       ForgeModuleFile.new(configs).render
     end
 
-    def self.sync(template, target_name)
+    def self.sync(template, target_name, mode)
```
ekohl: I'd prefer `mode=nil` and adding an `if` guard to `File.chmod`

lib/modulesync.rb
```diff
@@ -84,10 +84,16 @@ def self.manage_file(filename, settings, options)
       Renderer.remove(module_file(options[:project_root], module_name, filename))
     else
       templatename = local_file(options[:configs], filename)
+      template_file = if !File.exist?("#{templatename}.erb") && File.exist?(templatename)
```
ekohl: It's a bit ugly that we have to duplicate this logic from the renderer. I wonder if we should rip that out and do a major version bump.

lib/modulesync/renderer.rb
arjenz: I just pushed an attempt to do this.

lib/modulesync.rb
arjenz: I agree it's ugly, I added it only after I saw failing tests :)

I have no strong opinion on ripping it out, but I do think the whole ".erb extension is optional" and doing weird things like this to work around that in the code is rather confusing.

ekohl: I think it's sort of halfway in the migration. The original had no `.erb` extensions. Then it became the normal way and 'naked' filenames were deprecated. The migration was never completed, but the idea was to copy 'naked' filenames without templating them. IMHO we complete it and do a major version bump.

steinbrueckri:
👍

bastelfreak:
@ekohl can you have a look again please?

saz:
@bastelfreak @ekohl Any news on this one? Anything I can do to get this merged?

ekohl:
I forgot about this and now it has merge conflicts. I rebased it in https://github.com/voxpupuli/modulesync/pull/257. It moves around some code to avoid duplication.

PR #149, opened 2018-05-02
Add the ability to automatically merge downstream PRs created by modulesync

joestump:
## What does this PR do?

* Introduces a `manage_pr` function. Rubocop was (rightfully) complaining about the complexity of this block and brings the PR code a bit more in line with the existing `manage_${thing}` pattern.
* Introduces `--pr-auto-merge` to automatically merge downstream PRs created by Modulesync. 
* Excludes `spec/` from Rubocop checks.
* Switches to `$stdout` and `$stderr`, which worked best with RSpec's `output()` checks.

## Why is this change being made?

My team and I are so lazy we can't be bothered to manually merge all of the PRs our robots make. This lets our robots do it for us. 🤖😂 

## How was this tested? How can the reviewer verify your testing?

We're still testing this. Please consider this a work in progress until I've confirmed later.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,6 +7,7 @@ AllCops:
     - 'tmp/**/*'
     - 'pkg/**/*'
     - 'lib/monkey_patches.rb'
+    - 'spec/**/*'
 
 Style/HashSyntax:
   Enabled: false
--- a/README.md
+++ b/README.md
@@ -193,6 +193,9 @@ include:
 * Assign labels to the PR with `--pr-labels` or in `modulesync.yml` with the
   `pr_labels` attribute. **NOTE:** `pr_labels` should be a list. When using
   the `--pr-labels` CLI option, you should use a comma separated list.
+* `--pr-auto-merge` will attempt to automatically merge downstream PRs created
+  by modulesync. Skips PRs that aren't able to be merged when the 
+  `--skip-broken` option is passed.
 
 You can optionally set the `GITHUB_BASE_URL` environment variable to use GitHub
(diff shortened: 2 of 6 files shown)


ekohl:
@joestump could you have a look at my comment?

joestump:
@ekohl I can add checks for CI via the [statuses API](https://developer.github.com/v3/repos/statuses/#get-the-combined-status-for-a-specific-ref). We have branch protections on `master` for most/all of our repos. I guess we could let our robots push to `master` though.

bittner:
I feel this is a valuable change that makes sense. It's optional to activate, so we don't break existing functionality. :+1: 

@joestump, would it be hard to add a few tests for your implementation? Tests would be great!

I would merge the changes afterwards unless someone wants to veto against it.

joestump:
@bittner I'm not super great with Ruby testing. If you could point me at a PR I can crib from, I'll take a look at mocking something up.

bastelfreak:
Hi. Thanks for this awesome PR!
The Vox Pupuli release workflow currently looks like this:
A user runs a rake task locally which will create a git tag and a new commit on master. This will be pushed. Based on that we cannot set our branches on github.com to protected. The option to honour the status API would be very important. Is that something you could implement as well?

joestump:
@bittner tests added. @bastelfreak would it be possible to follow up with that functionality in a later PR? How should that function? Should it simply skip? Poll while waiting on status? Other?

ekohl:
I feel this feature doesn't fit within the tool and that makes me slightly against it. I'd prefer a tool that you can feed PRs, monitor them and merge when green. To me that's more the unix philosophy.

Note Gitlab has this built in but GH needs some special attention.

joestump:
Going to close this out. Relevant tests have been copied over to #171.

PR #152, opened 2018-05-27
Use namespace in directory structure when cloning repositories

### LINKED ISSUE #150: Custom namespace breaks when repos have matching names
With custom namespaces (PR #62) it's possible that you list equally-named repositories from different namespaces in your `managed_modules.yml` file, e.g.

```yaml
---
- module_foobar
- custom/module_foobar
```
However, in this case **only the very first** repository in this list gets updated. The flow of execution goes like this:

1. First, repository `<:git_base>:<:namespace>/module_foobar` is cloned into `<current_directory>/modules/module_foobar/`, modified and changes committed and pushed back to `origin`.
1. Then, repository `<:git_base>:custom/module_foobar` should be cloned, but modulesync [finds the existing directory](https://github.com/voxpupuli/modulesync/blob/master/lib/modulesync/git.rb#L59-L63), fetches changes, finds an updated repo and stops happily. **Read:** The custom-namespace repo is never cloned!

Example output:
```
$ msync update
No config file under ./config_defaults.yml found, using default values
Syncing module_foobar
Cloning repository fresh
Cloning from git@gitlab.example.com:default/module_foobar.git
Using repository's default branch: master
Syncing custom/module_foobar
Overriding any local changes to repositories in modules
Using repository's default branch: master
There were no files to update in module_foobar. Not committing.
```

We have to either remove the locally cloned repository (and clone into a directory with the same name again) or clone into directories named `./modules/<:namespace>/<:repo_name>` in general. I would prefer the latter.

Any comments before I start a PR?

bittner:
PR #62 added custom namespaces, but (unknowingly?) repository names had to be different across namespaces. A dangerous because almost entirely silent error, which easily gets overlooked. Details are described in issue #150.

These changes fix this limitation by cloning repositories into a `./modules/<namespace>/<repository>/` directory structure (instead of only `./modules/<repository>/` as of today).

Fixes #150 

*Sponsored by @vshn, The DevOps Company*

### DIFF
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -32,8 +32,8 @@ module ModuleSync
     "#{config_path}/#{MODULE_FILES_DIR}/#{file}"
   end
 
-  def self.module_file(project_root, puppet_module, file)
-    "#{project_root}/#{puppet_module}/#{file}"
+  def self.module_file(project_root, namespace, puppet_module, file)
+    "#{project_root}/#{namespace}/#{puppet_module}/#{file}"
   end
 
   def self.local_files(path)
@@ -78,16 +78,17 @@ module ModuleSync
   end
 
   def self.manage_file(filename, settings, options)
+    namespace = settings.additional_settings[:namespace]
     module_name = settings.additional_settings[:puppet_module]
     configs = settings.build_file_configs(filename)
     if configs['delete']
-      Renderer.remove(module_file(options[:project_root], module_name, filename))
+      Renderer.remove(module_file(options[:project_root], namespace, module_name, filename))
     else
       templatename = local_file(options[:configs], filename)
       begin
         erb = Renderer.build(templatename)
         template = Renderer.render(erb, configs)
-        Renderer.sync(template, module_file(options[:project_root], module_name, filename))
+        Renderer.sync(template, module_file(options[:project_root], namespace, module_name, filename))
       rescue # rubocop:disable Lint/RescueWithoutErrorClass
         STDERR.puts "Error while rendering #{filename}"
         raise
@@ -101,20 +102,18 @@ module ModuleSync
       raise unless options[:skip_broken]
     end
 
-    puts "Syncing #{puppet_module}"
     namespace, module_name = module_name(puppet_module, options[:namespace])
+    git_repo = "#{namespace}/#{module_name}"
     unless options[:offline]
-      git_base = options[:git_base]
-      git_uri = "#{git_base}#{namespace}"
-      Git.pull(git_uri, module_name, options[:branch], options[:project_root], module_options || {})
+      Git.pull(options[:git_base], git_repo, options[:branch], options[:project_root], module_options || {})
     end
-    module_configs = Util.parse_config("#{options[:project_root]}/#{module_name}/#{MODULE_CONF_FILE}")
+    module_configs = Util.parse_config("#{options[:project_root]}/#{namespace}/#{module_name}/#{MODULE_CONF_FILE}")
     settings = Settings.new(defaults[GLOBAL_DEFAULTS_KEY] || {},
                             defaults,
                             module_configs[GLOBAL_DEFAULTS_KEY] || {},
                             module_configs,
                             :puppet_module => module_name,
-                            :git_base => git_base,
+                            :git_base => options[:git_base],
                             :namespace => namespace)
     settings.unmanaged_files(module_files).each do |filename|
       puts "Not managing #{filename} in #{module_name}"
@@ -124,10 +123,10 @@ module ModuleSync
     files_to_manage.each { |filename| manage_file(filename, settings, options) }
 
     if options[:noop]
-      Git.update_noop(module_name, options)
+      Git.update_noop(git_repo, options)
     elsif !options[:offline]
       # Git.update() returns a boolean: true if files were pushed, false if not.
-      pushed = Git.update(module_name, files_to_manage, options)
+      pushed = Git.update(git_repo, files_to_manage, options)
       return nil unless pushed && options[:pr]
 
       # We only do GitHub PR work if the GITHUB_TOKEN variable is set in the environment.
--- a/lib/modulesync/git.rb
+++ b/lib/modulesync/git.rb
@@ -46,12 +46,13 @@ module ModuleSync
     end
 
     def self.pull(git_base, name, branch, project_root, opts)
+      puts "Syncing #{name}"
       Dir.mkdir(project_root) unless Dir.exist?(project_root)
 
       # Repo needs to be cloned in the cwd
       if !Dir.exist?("#{project_root}/#{name}") || !Dir.exist?("#{project_root}/#{name}/.git")
         puts 'Cloning repository fresh'
-        remote = opts[:remote] || (git_base.start_with?('file://') ? "#{git_base}/#{name}" : "#{git_base}/#{name}.git")
+        remote = opts[:remote] || (git_base.start_with?('file://') ? "#{git_base}#{name}" : "#{git_base}#{name}.git")
         local = "#{project_root}/#{name}"
         puts "Cloning from #{remote}"
         repo = ::Git.clone(remote, local)
--- a/features/update.feature
+++ b/features/update.feature
@@ -31,7 +31,7 @@ Feature: update
       Files added:
       test
       """
-    Given I run `cat modules/puppet-test/test`
+    Given I run `cat modules/maestrodev/puppet-test/test`
     Then the output should contain "aruba"
 
   Scenario: Using skip_broken option and adding a new file to repo without write access
@@ -120,7 +120,7 @@ Feature: update
       Files added:
       test
       """
-    Given I run `cat modules/puppet-test/test`
+    Given I run `cat modules/maestrodev/puppet-test/test`
     Then the output should contain "aruba"
 
   Scenario: Adding a new file using global values
@@ -153,7 +153,7 @@ Feature: update
       Files added:
       test
       """
-    Given I run `cat modules/puppet-test/test`
+    Given I run `cat modules/maestrodev/puppet-test/test`
     Then the output should contain "aruba"
 
   Scenario: Adding a new file overriding global values
@@ -189,7 +189,7 @@ Feature: update
       Files added:
       test
       """
-    Given I run `cat modules/puppet-test/test`
+    Given I run `cat modules/maestrodev/puppet-test/test`
     Then the output should contain "aruba"
 
   Scenario: Adding a new file ignoring global values
@@ -225,7 +225,7 @@ Feature: update
       Files added:
       test
       """
-    Given I run `cat modules/puppet-test/test`
+    Given I run `cat modules/maestrodev/puppet-test/test`
     Then the output should contain "aruba"
 
   Scenario: Adding a file that ERB can't parse
@@ -342,7 +342,7 @@ Feature: update
       Files changed:
       +diff --git a/Gemfile b/Gemfile
       """
-    Given I run `cat modules/puppet-test/Gemfile`
+    Given I run `cat modules/maestrodev/puppet-test/Gemfile`
     Then the output should contain:
       """
       source 'https://somehost.com'
@@ -405,7 +405,7 @@ Feature: update
       Not managing Gemfile in puppet-test
       """
     And the exit status should be 0
-    Given I run `cat modules/puppet-test/Gemfile`
+    Given I run `cat modules/maestrodev/puppet-test/Gemfile`
     Then the output should contain:
       """
       source 'https://rubygems.org'
@@ -488,8 +488,8 @@ Feature: update
       """
       some spec_helper fud
       """
-    And a directory named "modules/puppetlabs-apache/spec"
-    And a file named "modules/puppetlabs-apache/spec/spec_helper.rb" with:
+    And a directory named "modules/puppetlabs/puppetlabs-apache/spec"
+    And a file named "modules/puppetlabs/puppetlabs-apache/spec/spec_helper.rb" with:
       """
       This is a fake spec_helper!
       """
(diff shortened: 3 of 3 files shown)


bittner:
Ruby version 2.0.0 was [dropped almost  2 years ago](https://www.ruby-lang.org/en/news/2016/02/24/support-plan-of-ruby-2-0-0-and-2-1/) and Ruby 2.5.1 was [released this March](https://www.ruby-lang.org/en/news/2018/03/28/ruby-2-5-1-released/) (and is included in the latest LTS version of Ubuntu). Also the other versions have advanced, in all cases with security fixes.

Should I update the [.travis.yml](https://github.com/voxpupuli/modulesync/blob/master/.travis.yml) file accordingly in this PR?

ekohl:
I think this change makes sense but makes me wonder if we should do a major version bump. Dropping Ruby 2.0 would also be a major bump IMHO so that's nice to combine.

> Should I update the .travis.yml file accordingly in this PR?

No, that's a different PR. Dropping 2.0 can be fine but is a larger change that would clutter it. It's also clearer when generating a changelog.

bittner:
I have no problem with a version bump, especially when it [jumps to 1.0.0](https://github.com/voxpupuli/modulesync/blob/master/modulesync.gemspec#L6). :tada: :smiley: 

Keep in mind, however, that the changes from this PR are actually only a bugfix. The structural change happens, technically speaking, under the hood. (Who relies on and messes with the files and folders of the dynamically cloned repositories is asking for trouble, don't we agree?)

It's neither a new feature nor a breaking change that I introduce. This would only legitimate a patch version number bump according to Semver. Again though, I don't really care.

Bottom line: Is there anything I should change or add to make this PR get merged?

bittner:
Anything missing before this PR can be merged?

ekohl:
While it's not an official API, we do have scripts that assume the `modules/*` structure rather than `modules/*/*`. See https://github.com/voxpupuli/modulesync_config/tree/master/bin for example. One could argue this functionality should be in modulesync itself (like cleaning).

One option I'm thinking about is to not introduce the extra layer if there's just 1 namespace, but that also creates some inconsistent behavior. Any thoughts?

bittner:
> While it's not an official API, we do have scripts that assume the modules/* structure

Doesn't this sound a bit like "we have a feature that exploits this bug"? -- It's a bug. So, we also need to fix those scripts. If it's not easy, because "the behavior is relied on heavily" and "outside this project", then we may need to introduce a deprecation timeline or something similar.

> See https://github.com/voxpupuli/modulesync_config/tree/master/bin for example. 

* [ ] [bin/clean-git-checkouts](https://github.com/voxpupuli/modulesync_config/blob/master/bin/clean-git-checkouts#L5) -- that looks easy to fix
* [ ] [bin/clean-metadata](https://github.com/voxpupuli/modulesync_config/blob/master/bin/clean-metadata#L20) -- that looks easy to fix
* [ ] [bin/count-unreleased-commits](https://github.com/voxpupuli/modulesync_config/blob/master/bin/count-unreleased-commits#L5) -- that looks easy to fix
* [ ] [bin/create-pull-requests](https://github.com/voxpupuli/modulesync_config/blob/master/bin/create-pull-requests#L10) -- that looks easy to fix
* [ ] [bin/get_all_the_diffs](https://github.com/voxpupuli/modulesync_config/blob/master/bin/get_all_the_diffs#L36) -- that file may not need any changes (not sure though)

Should I go ahead?  Are there tests for them (that allow safe refactoring)?

> not introduce the extra layer if there's just 1 namespace, but that also creates some inconsistent behavior.

It's probably easier and cleaner to keep things simple and fix the issue with all its dependencies properly.

ekohl:
> Doesn't this sound a bit like "we have a feature that exploits this bug"? -- It's a bug. So, we also need to fix those scripts. If it's not easy, because "the behavior is relied on heavily" and "outside this project", then we may need to introduce a deprecation timeline or something similar.

They were written before the multi-namespace feature I think. That's why I suggested a major version bump so you can easily depend on the old or the new version. Note the voxpupuli config is just one. I don't know how many others have scripted around modulesync.

There's another approach: we can implement a command in modulesync to print a list of directories to iterate. That way you provide a more formal way to script against and it really becomes an internal detail.

bittner:
> That's why I suggested a major version bump so you can easily depend on the old or the new version.

Fine with me.

> we can implement a command in modulesync to print a list of directories to iterate.

I would walk the most straight-forward route to start with, and only align the scripts mentioned above with the changes to ModuleSync in this PR. They would be small changes, which can be improved upon in a later iteration (with the approaches you mention).

Is that acceptable for you?

bittner:
There [doesn't seem to be a way](https://github.com/voxpupuli/modulesync/blob/master/lib/modulesync/cli.rb#L59-L123) to find out the ModuleSync version (both `msync --version` and `msync version` don't cooperate, and also `msync --help` doesn't give any hint on the version being in use). This makes it hard to make safe, backwards-compatible changes to the [scripts in modulesync_config](https://github.com/voxpupuli/modulesync_config/tree/master/bin), I'm afraid. :worried: 

In other words, regardless to whether we take the "print a list of directories" or the "straight-forward route" approach we'll end up with scripts in the `modulesync_config` repository that require an updated version of ModuleSync. -- Well, not totally true! We could add a `--version` option now and use that in the scripts. But still, a bit quirky. Any thoughts?

I still would prefer to get this PR merged (and the related bug fixed). This is a bit of a road blocker, because it's actually a dangerous features (since the [failure remains unnoticed](https://github.com/voxpupuli/modulesync/issues/150)).

ekohl:
@bittner with a major version bump users can require the new version in their Gemfiles. So in voxpupuli we could modify the scripts for the new version and add a `>= x` in `Gemfile`. I'd be fine with that.

bittner:
I could bump the version to 1.0.0 in this PR, not a big deal. Would the PR them get merged?

bittner:
**ping** :bell: 

Anyone wanting to merge this bugfix?

rnelson0:
@ekohl Since you've looked at the modulesync_config repo, would we have a problem merging it if someone released a new gem afterward, even if it was 1.0.0? I just want to make sure it's properly bound there so we don't cause problems due to bad timing of the changes between the two repos.

bittner:
**ping** :bell: @ekohl 

I can apply the changes to the config repo right afterwards. Let's merge this. It's really a fix that makes sense.

lib/modulesync.rb
```diff
@@ -101,20 +102,18 @@ def self.manage_module(puppet_module, module_files, module_options, defaults, op
       raise unless options[:skip_broken]
     end
 
-    puts "Syncing #{puppet_module}"
     namespace, module_name = module_name(puppet_module, options[:namespace])
+    git_repo = "#{namespace}/#{module_name}"
```
ekohl: I don't think this is correct. IMHO you should introduce a new variable `namespace_root` that is the project root + namespace instead.

lib/modulesync.rb
```diff
@@ -78,16 +78,17 @@ def self.hook(options)
   end
 
   def self.manage_file(filename, settings, options)
+    namespace = settings.additional_settings[:namespace]
```
ekohl: I'm not sure this is correct. If a module is overridden then `module_name` extracts it from the module name rather than the `namespace` setting. Wouldn't this break if `namespace/module` is used?

lib/modulesync.rb
bittner: Is this just a matter of taste, or a technical issue?

> new variable `namespace_root` that is the project root + namespace

What do you intend with "project root"? Not the repository? Can you make a concrete example?

P.S.: You can write, _"Replace this by ..."_. I promise, I won't be offended.

lib/modulesync.rb
```diff
@@ -124,10 +123,10 @@ def self.manage_module(puppet_module, module_files, module_options, defaults, op
     files_to_manage.each { |filename| manage_file(filename, settings, options) }
 
     if options[:noop]
-      Git.update_noop(module_name, options)
+      Git.update_noop(git_repo, options)
     elsif !options[:offline]
       # Git.update() returns a boolean: true if files were pushed, false if not.
-      pushed = Git.update(module_name, files_to_manage, options)
+      pushed = Git.update(git_repo, files_to_manage, options)
```
bittner: @ekohl Interestingly, even though the tests pass this may turn out to be an issue. Because in [Git.update](https://github.com/vshn/modulesync/blob/bb5c88394105d0a5696c670db985932d5a410ea3/lib/modulesync/git.rb#L118) the `git_repo` value is prepended with a `:project_root` value. Not sure, though, honestly. (Ruby is not my first language.)

lib/modulesync.rb
ekohl: > Is this just a matter of taste, or a technical issue?

I think it should be that way because the code that's called (`Git.pull`) does a `Dir.mkdir(project_root) unless Dir.exist?(project_root)`. Your version changes to rely on `git clone` to create parent directories. My git (2.14.4) does this, but I don't know if all git versions do this. If we rely on git to create parent directories, we might as well remove the `Dir.mkdir` there. This is me trying to figure out what's going on because I don't know the project that well.

> What do you intend with "project root"? Not the repository? Can you make a concrete example?

I meant `namespace_root = File.join(options[:project_root], namespace)` (though I'm not attached to the variable name).

> P.S.: You can write, "Replace this by ...". I promise, I won't be offended.

That sounds like I know what I'm doing :P

lib/modulesync.rb
bittner: I think it does. Can you imagine a test case that would make the code fail?

lib/modulesync.rb
bittner: I think this problem is artificial. When we look at the [Git.pull](https://github.com/voxpupuli/modulesync/blob/master/lib/modulesync/git.rb#L48-L52) code we see that the `Dir.mkdir` is needed. But not for the Git operation, but because we want to do Git things _in_ that directory.

We can still try to prove this is false and will fail. If we manage to come up with a test case that fails. I think that's the route that makes really sense. (I rely a lot on tests, especially for languages that I don't fully master. The existing ones have helped me to get as far as I am now.)

lib/modulesync.rb
bittner: This makes sense now, chiming in with [my discovery above](https://github.com/voxpupuli/modulesync/pull/152#discussion_r201506920): The `:project_root` has nothing to do with Git. It's just the place where all Git operations are meant to be performed in.

The `Git.update` only does things in the final _Git_ project root = `:project_root` + Git repository path (even if this is two levels).

bittner:
Hello from hot Zurich city center! :hotsprings: :sun_with_face: 

Where did we leave off? Can this PR be merged as it is? (I hope it can. Otherwise let me know what should look like how, please.)

bittner:
Are we ready to merge this PR? Remember, it fixes a bug.

:bell: ***ping***

bastelfreak:
@ekohl can you take a look at this?

bittner:
Any chance this bugfix is getting merged?

bittner:
Rebased on latest `master/HEAD`.

lib/modulesync.rb
```diff
@@ -101,20 +102,18 @@ def self.manage_module(puppet_module, module_files, module_options, defaults, op
...
     namespace, module_name = module_name(puppet_module, options[:namespace])
+    git_repo = "#{namespace}/#{module_name}"
     unless options[:offline]
-      git_base = options[:git_base]
-      git_uri = "#{git_base}#{namespace}"
-      Git.pull(git_uri, module_name, options[:branch], options[:project_root], module_options || {})
+      Git.pull(options[:git_base], git_repo, options[:branch], options[:project_root], module_options || {})
     end
-    module_configs = Util.parse_config("#{options[:project_root]}/#{module_name}/#{MODULE_CONF_FILE}")
+    module_configs = Util.parse_config("#{options[:project_root]}/#{namespace}/#{module_name}/#{MODULE_CONF_FILE}")
```
ekohl: This should use `git_repo` to stay consistent

lib/modulesync/git.rb
```diff
@@ -46,12 +46,13 @@ def self.switch_branch(repo, branch)
     end
 
     def self.pull(git_base, name, branch, project_root, opts)
+      puts "Syncing #{name}"
       Dir.mkdir(project_root) unless Dir.exist?(project_root)
```
ekohl: This can be dropped if you're going to pass in name as `dir/subdir` because you're going to rely on git to create a directory anyway.

bittner:
@mhutter Can we look into this together and refactor it properly, so that the objections raised by @ekohl get addressed? My Ruby skills are way too basic to do this properly. There shouldn't be missing too much, though.

mhutter:
@ekohl I see your points and agree that some of the modules need some refactoring. For example the current implementation of the `Git` module has knowledge of concepts like namsepaces, projects and so on while all it is supposed to do os interacting with git (which only requires knowledge of repository URLs, file system paths and possibly branch names). This would maybe make things like #159 easier as well.

However, this is out of scope of this PR. The goal of this PR is to fix the functional defect of modulesync described in #150, while changing as little on the overall codebase as needed.

bittner:
@ekohl Can we merge this PR as suggested by @mhutter? It fixes an actual bug. Refactoring the Git module would then be the job of a separate PR.

bittner:
Is anyone else from the project maintainers interested in merging this PR?

I see that @ekohl is very involved with The Foreman. I can understand that this repo is "just a distraction". Maybe someone of you @voxpupuli people can help?

I may be repeating myself, but this PR fixes an actual bug. A bug that is easily overlooked in ops business ([see description](https://github.com/voxpupuli/modulesync/issues/150)). :bomb: :worried:

bittner:
Who can look at the changes in this PR and merge them---or decide to refuse them forever?

@bastelfreak, @rnelson0, @ekohl, someone else?

I think, we can easily define follow-up tasks to implement improvements on shortcomings of the current modules and scripts that have become apparent now. And that should be done separately to keep the change set small enough.

bittner:
:bell: ***ping***

bittner:
Thank you for merging! :1st_place_medal:

bittner:
It makes sense to release a new version now, for type of bug fix. It may make sense to do a major version number jump. Though, strictly speaking, it's an internal change. The external behavior doesn't change.

PR #157, opened 2018-08-31
Fix opts[:remote] call when using hashes for configs

joestump:
When you use a [hash instead of a list](https://github.com/voxpupuli/modulesync/blob/master/lib/modulesync.rb#L164) in `managed_modules.yml` the keys in `module_options` are strings and not symbols. This results in [`opts[:remote]`](https://github.com/voxpupuli/modulesync/blob/master/lib/modulesync/git.rb#L54) being ignored.

I ran across this when using AWS CodeCommit, which doesn't use `.git` suffixes on their URIs. The following `managed_modules.yml` works with this fix:

```yaml
---
my-fancy-repo:
  remote: https://git-codecommit.us-west-2.amazonaws.com/v1/repos/my-fancy-repo
```


ekohl:
Could you add a test for this?

raphink:
Fixed in #185

PR #158, opened 2018-10-04
Add -t option to populate managed modules using GitHub's repository topics

joestump:
GitHub supports [tagging repositories with "topics"](https://developer.github.com/v3/repos/#list-organization-repositories). This PR leverages this functionality to populate a list of managed modules rather than having to manually maintain `managed_modules.yml`.

This allows teams to add the topic `ruby` to a GitHub repository and then use the `-t ruby` option in their Ruby Modulesync CI/CD job to grab all of the repositories to sync.

### DIFF
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -11,10 +11,16 @@ require 'modulesync/util'
 require 'monkey_patches'
 
 GITHUB_TOKEN = ENV.fetch('GITHUB_TOKEN', '')
+GITHUB_ORGANIZATION = ENV.fetch('GITHUB_ORGANIZATION', '')
 
 Octokit.configure do |c|
   c.api_endpoint = ENV.fetch('GITHUB_BASE_URL', 'https://api.github.com')
+  c.auto_paginate = true
+  c.default_media_type = ::Octokit::Preview::PREVIEW_TYPES[:topics]
 end
+
+GITHUB = Octokit::Client.new(:access_token => GITHUB_TOKEN)
+
 
 module ModuleSync
   include Constants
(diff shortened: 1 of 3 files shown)


lib/modulesync.rb
```diff
@@ -50,8 +58,16 @@ def self.module_files(local_files, path)
     local_files.map { |file| file.sub(/#{path}/, '') }
   end
 
-  def self.managed_modules(config_file, filter, negative_filter)
-    managed_modules = Util.parse_config(config_file)
+  def self.managed_modules(config_file, filter, negative_filter, topic)
+    if topic
+      managed_modules = []
+      GITHUB.org_repos(GITHUB_ORGANIZATION).each do |repo|
```
ekohl: I'm wondering if `managed_modules = GITHUB.org_repos(GITHUB_ORGANIZATION).filter { |r| r.topics.include?(topic) }.map(&:name)` is easier to read

lib/modulesync.rb
```diff
@@ -11,11 +11,19 @@
...
 Octokit.configure do |c|
   c.api_endpoint = ENV.fetch('GITHUB_BASE_URL', 'https://api.github.com')
+  c.auto_paginate = true
+
+  # We use a different Accept header by default here so we can get access
+  # to the developer's preview that enables repository topics.
+  c.default_media_type = ::Octokit::Preview::PREVIEW_TYPES[:topics]
 end
 
+GITHUB = Octokit::Client.new(:access_token => GITHUB_TOKEN)
```
ekohl: I'm not sure about a global with state. Perhaps it should be a lazy (static) method on the module. This can help in case it's used in an offline context.

bittner:
Interesting! This sounds like driven by the same idea we had for managing GitLab projects with their projects topics. :+1: 

See our Python CLI package [concierge-cli](https://pypi.org/project/concierge-cli/) on PyPI for reference.

PR #163, opened 2018-12-27
release 0.10.0

bastelfreak:


bittner:
Here on GitHub the releases 0.9.0 and 0.10.0 didn't show up cleanly; only the tags were visible. That was a bit confusing. Hence, I created the "Releases" (manually). I hope that's fine.

PR #166, opened 2019-07-05
Small improvements and fixes

vStone:
I started digging into the code and trying to understand I refactored some things. These changes may assist others with similar endeavors. 

* Bring back max line length for rubocop
* use File.join() where applicable
* reuse module_file() for getting the `MODULE_CONF_FILE` (`.sync.yml`)
* renamed local_files() -> find_template_files() because it is what it does.
* rename module_files() to relative_names() because that is what it does.
* rename local variable `path` in update() to `local_template_dir` because that is what it is.


vStone:
*woops*

PR #169, opened 2019-09-16
Fix overwriting of :global values

anothertobi:
When defining `:global` values in `.sync.yml` we faced the issue, that they were not accessible via `@configs` in the `.erb` templates.

Further debugging showed that the values were not loaded into `@configs`.

Looking at the code we found out that the values for `module_defaults` are already resolved in `modulesync.rb` (like the `global_defaults`). Therefore the additional `lookup_config` basically discards all values configured in `.sync.yml` files under `:global` as it cannot find something under the `target_name` key.

### DIFF
--- a/features/update.feature
+++ b/features/update.feature
@@ -756,6 +756,9 @@ Feature: update
     And a file named "config_defaults.yml" with:
       """
       ---
+      :global:
+        global-default: some-default
+        global-to-overwrite: to-be-overwritten
       spec/spec_helper.rb:
         require:
           - puppetlabs_spec_helper/module_helper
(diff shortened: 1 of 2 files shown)


bittner:
Is there a test that could reproduce the issue? One that would prove that the issue existed, and that it's fixed with those changes.

Ideally, you push a commit with a (new or updated) test first, and then push the changes that fix the issue, afterwards. This way we could have Travis document the fix.

Can you do this?

srueg:
Failing build (introducing only the tests): https://travis-ci.org/voxpupuli/modulesync/builds/585659772
Successful build (including the fix): https://travis-ci.org/voxpupuli/modulesync/builds/585660755

bittner:
Looks good! Thank you for the extra mile! :1st_place_medal:

PR #170, opened 2019-09-19
Release 1.0.0

srueg:
We propose to make the jump to 1.0.0 since https://github.com/voxpupuli/modulesync/pull/152 introduced a substantial change in internal behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## 2019-09-19 - 1.0.0
+
+This is the first stable release! 🎉
+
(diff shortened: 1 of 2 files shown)


bittner:
Looks good to me. The major version bump makes sense, which was also and originally suggested by @ekohl in https://github.com/voxpupuli/modulesync/pull/152#issuecomment-392319507.

Apart from that, the higher number really doesn't do any harm and depicts somewhat an achieved maturity of the project (which, given the apparently high number of users, can be taken as a given).

Any other thoughts, anyone?

PR #173, opened 2020-04-05
Set the pr_labels option type to array

ekohl:


ekohl:
I think this failure already existed in master and only happens on Ruby 2.4. I think that may pull in a newer Thor.
```
    When I run `msync update -m 'deletes a file that doesnt exist!' -f puppet-test` # vendor/bundle/ruby/2.4.0/gems/aruba-1.0.0/lib/aruba/cucumber/command.rb:3
    And the exit status should be 0                                                 # vendor/bundle/ruby/2.4.0/gems/aruba-1.0.0/lib/aruba/cucumber/command.rb:262
      expected that command "msync update -m 'deletes a file that doesnt exist!' -f puppet-test" has exit status of "0", but has "1". (RSpec::Expectations::ExpectationNotMetError)
      ./vendor/bundle/ruby/2.4.0/bin/ruby_executable_hooks:15:in `eval'
      ./vendor/bundle/ruby/2.4.0/bin/ruby_executable_hooks:15:in `<main>'
      features/update.feature:466:in `And the exit status should be 0'
```

ekohl:
Looks like the difference is in aruba 0.14.14 vs 1.0.0

Issue #80, opened 2016-06-24
create a script/extension to automatically create PRs


bastelfreak:
I think the title describes it very good. also @bbriggs said he is interested in implementing this

bbriggs:
Unless anyone has objections, I plan to just extend the existing gem in Ruby as opposed to adding a new script. The feature will be triggered with a command line switch.

bastelfreak:
@danzilio do you have any suggestions here?

danzilio:
Sounds good to me. Honestly, I'd like to see a plugin system here (for example, it would be nice to support submitting PRs for GitLab and Stash as well) but that's a much bigger task. Let's see what this looks like and go from there.

bbriggs:
So maybe a modulesync-github-pr gem, then?

bbriggs:
And then have the modulesync-*-pr gems provide a common API to modulesync so it can choose the appropriate gem/plugin to call when --auto-pr is enabled?

danzilio:
@bbriggs Exactly! I'd loooove to see that. But, right now, the API in modulesync isn't particularly robust. It would definitely be an undertaking. Although, maybe this is a good place to start...

rnelson0:
modulesync-pr-* would probably be better for sorting, tab completion, etc,

>

-- 

Rob Nelson
rnelson0@gmail.com

bbriggs:
I guess now I get to write my first gem. Yay Ruby!

sean797:
@bbriggs did you ever make a start on this ?

bbriggs:
:(

bbriggs:
In my heart I really want to but realistically I don't think I'll be able to get to this.

timhughes:
not the most elegant code but here is what I use for gitlab  https://gist.github.com/timhughes/80928f02baf7ee5c52180b16f0a2ba0b

ekohl:
Personally I use the following:

```bash
rm -rf modules
bundle exec msync update --noop
# Ensure all modules are good to go - hopefully nothing was changed in files but not .sync.yml
MESSAGE="..."
for mod in modules/* ; do
    echo "Handling $mod"
    git commit -a -m "$MESSAGE" && git push $(whoami) HEAD -u && hub pull-request -m "$MESSAGE"
done
```

Personally I like the control over msync doing the commits but in general it's clear. https://hub.github.com/ is a great tool and I'd be a lot less efficient at using github if I had to use the UI to do PRs.

[...]

PR #175, opened 2020-05-18
Add support for GitLab merge requests (MRs)

chamer81:
Adds an option to create a gitlab MR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,10 @@
-tmp/
-modules/
 *.gem
-Gemfile.lock
 .bundle/
+.ruby-version
+.vscode/
+Gemfile.lock
+bin/bundle
+bin/rspec
+modules/
+tmp/
+vendor/
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -1,4 +1,5 @@
 require 'fileutils'
+require 'gitlab'
 require 'octokit'
 require 'pathname'
 require 'modulesync/cli'
@@ -11,9 +12,14 @@ require 'modulesync/util'
 require 'monkey_patches'
 
 GITHUB_TOKEN = ENV.fetch('GITHUB_TOKEN', '')
+GITLAB_TOKEN = ENV.fetch('GITLAB_TOKEN', '')
 
 Octokit.configure do |c|
   c.api_endpoint = ENV.fetch('GITHUB_BASE_URL', 'https://api.github.com')
+end
+
+Gitlab.configure do |c|
+  c.endpoint = ENV.fetch('GITLAB_BASE_URL', 'https://gitlab.com/api/v4')
 end
 
 module ModuleSync # rubocop:disable Metrics/ModuleLength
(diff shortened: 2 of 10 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
```
ekohl, line 1: Please keep this out

bin/bundle
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env ruby
```
ekohl, line 1: Please keep this out

bin/rspec
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env ruby
```
ekohl, line 1: Please keep this out

.rubocop_todo.yml
```diff
@@ -1,70 +0,0 @@
-# This configuration was generated by
```
ekohl, line 1: Could you submit this as a PR first? It makes reviewing this PR easier.

lib/modulesync.rb
```diff
@@ -1,4 +1,5 @@
 require 'fileutils'
+require 'gitlab'
```
ekohl, line 2: If this becomes a hard dependency then it needs to be in the gemspec

bittner:
Thank you @ekohl for the quick feedback. We'll take that immediately into account. We're still at fixing some rough edges of the implementation and doing internal review.

bittner:
Relates to https://github.com/voxpupuli/modulesync/issues/80#issuecomment-391329471.

bittner:
We have cleaned up the commits and I've consolidated the changes into a single commit.

There are still a few topics open: (at least from a first, quick glance)

- Apart from GitLab.com, we need to allow GitLab support for **private instances** ([`config.endpoint`](https://github.com/narkoz/gitlab#usage) is set for this); presumably, this needs to be configurable
- The added / updated **tests** don't seem complete
- The **lazy-loading** of the GitHub and GitLab dependencies is still pending realization

.gitignore
```diff
@@ -1,5 +1,10 @@
...
 *.gem
 Gemfile.lock
 .bundle/
+.ruby-version
+.vscode/
+bin/bundle
+bin/rspec
+modules/
+tmp/
+vendor/
```
ekohl, line 10: This would typically be a separate commit.

Gemfile
```diff
@@ -2,5 +2,7 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 
 gemspec
 
+gem 'aruba'
```
ekohl, line 5: Why is this needed, looks redundant:
https://github.com/voxpupuli/modulesync/blob/7202e6bfc819a8e7323a03c9391cf45b2b6df9d2/modulesync.gemspec#L20

Gemfile
```diff
@@ -2,5 +2,7 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
...
 gem 'cucumber', '< 3.0' if RUBY_VERSION < '2.1'
+gem 'gitlab' # gem 'gitlab', github: 'NARKOZ/gitlab'
 gem 'octokit', '~> 4.9'
```
ekohl, line 8: This was there, but I can't explain why because this should be sufficent:
https://github.com/voxpupuli/modulesync/blob/7202e6bfc819a8e7323a03c9391cf45b2b6df9d2/modulesync.gemspec#L27

Gemfile
```diff
@@ -2,5 +2,7 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 
 gemspec
 
+gem 'aruba'
 gem 'cucumber', '< 3.0' if RUBY_VERSION < '2.1'
+gem 'gitlab' # gem 'gitlab', github: 'NARKOZ/gitlab'
```
ekohl, line 7: This is redundant since it's in the gemspec

mhutter:
Updated the first commit to clean up dependencies in `Gemfile` and gemspec. Added a second commit to refactor the existing code. This (as a side-effect) also implements lazy-loading and client reuse.

mhutter:
Still looking into that failing cucumber test, not sure how that's related...

mhutter:
One thing that caught my eye when moving code around: The Gitlab implementation will ensure the configured labels are added to the PR when the PR is created, but will NOT add any labels if there is already an open PR around.

The GitHub implementation will always update the existing PR - and fail horribly from looking at the code (note this was already broken before this PR)

@ekohl what is the intended behaviour?

bittner:
@ekohl Can you clarify?

lib/modulesync.rb
```diff
@@ -11,10 +10,7 @@
 require 'monkey_patches'
 
 GITHUB_TOKEN = ENV.fetch('GITHUB_TOKEN', '')
-
-Octokit.configure do |c|
-  c.api_endpoint = ENV.fetch('GITHUB_BASE_URL', 'https://api.github.com')
-end
+GITLAB_TOKEN = ENV.fetch('GITLAB_TOKEN', '')
```
ekohl, line 13: I think we can avoid these constants and only use them inline since they're only used in a single place.

mhutter: Good point, totally missed it. Will adjust accordingly

ekohl:
This looks good to me. It's still marked as a draft so let me know when you think it's ready to be merged.

mhutter:
I broke the check in my last commit, fixed.

I think the PR is ready, can you remove the "Draft" status @chamer81 ?

I'll open a new Issue to discuss the PR label behaviour. Fixing it should be trivial.

ekohl:
Thanks!

bittner:
Thanks everyone for the good collaboration! :1st_place_medal: :confetti_ball:

lib/modulesync.rb
```diff
@@ -201,4 +168,20 @@ def self.update(options)
...
+
+  def self.create_pr_manager
+    github_token = ENV.fetch('GITHUB_TOKEN', '')
+    gitlab_token = ENV.fetch('GITLAB_TOKEN', '')
+
+    if !github_token.empty?
+      require 'modulesync/pr/github'
+      ModuleSync::PR::GitHub.new(github_token, ENV.fetch('GITHUB_BASE_URL', 'https://api.github.com'))
+    elsif !gitlab_token.empty?
+      require 'modulesync/pr/github'
```
raphink, line 180: Should be `gitlab`

bittner: Fixed via commit 2b5388230fb39dff85e3522db753d20e8c88dfab.

PR #178, opened 2020-07-03
Allow to override PR parameters per module

raphink:


lib/modulesync.rb
```diff
@@ -177,7 +193,7 @@ def self.create_pr_manager
       require 'modulesync/pr/github'
       ModuleSync::PR::GitHub.new(github_token, ENV.fetch('GITHUB_BASE_URL', 'https://api.github.com'))
     elsif !gitlab_token.empty?
-      require 'modulesync/pr/github'
+      require 'modulesync/pr/gitlab'
```
bastelfreak, line 196: can you submit this as a bugfix PR?

PR #183, opened 2020-07-17
Allow absolute path for config files

raphink:
This PR makes it possible to pass configuration as absolute file names instead of relative.

This will allow to pass `managed_modules_conf` from a tempfile for example (to secure possible tokens in there).


raphink:
Sure

PR #184, opened 2020-07-20
Expose --managed_modules_conf

raphink:


raphink:
CI fails because of number of options in CLI... Do we want to ignore this?

PR #185, opened 2020-07-21
Symbolize keys in managed_modules except for module names

raphink:
This allows to use symbols or strings for subkeys

Properly fixes #167 and #157

### DIFF
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -171,7 +171,8 @@ module ModuleSync # rubocop:disable Metrics/ModuleLength
     # managed_modules is either an array or a hash
     managed_modules.each do |puppet_module, module_options|
       begin
-        manage_module(puppet_module, module_files, module_options, defaults, options)
+        mod_options = module_options.nil? ? nil : Util.symbolize_keys(module_options)
+        manage_module(puppet_module, module_files, mod_options, defaults, options)
       rescue # rubocop:disable Lint/RescueWithoutErrorClass
         $stderr.puts "Error while updating #{puppet_module}"
         raise unless options[:skip_broken]
--- a/README.md
+++ b/README.md
@@ -197,14 +197,14 @@ For GitHub Enterprise and self-hosted GitLab instances you also need to set the
 ```yaml
 ---
 repo1:
-  :github:
-    :token: 'EXAMPLE_TOKEN'
-    :base_url: 'https://api.github.com/'
+  github:
+    token: 'EXAMPLE_TOKEN'
+    base_url: 'https://api.github.com/'
 
 repo2:
-  :gitlab:
-    :token: 'EXAMPLE_TOKEN'
-    :base_url: 'https://git.example.com/api/v4'
+  gitlab:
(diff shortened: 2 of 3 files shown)


bastelfreak:
what's the benefit of using symbols? Is that something that should be added to the README.md with an example/explanation?

raphink:
@bastelfreak right now, only symbols are possible in the config file, and I'd like to use strings instead (because `gopass` uses `yamlpath`, which doesn' support `:` in keys at the moment).

Generally in Ruby, symbols are used for hash keys because they're more efficient than strings for static data (two identical symbols only point to one location in memory), but they're cumbersome to work with, esp. when you don't really care about high perfs.

raphink:
@ekohl yes, because it's backwards-compatible.

lib/modulesync.rb
```diff
@@ -171,6 +171,7 @@ def self.update(options)
     # managed_modules is either an array or a hash
     managed_modules.each do |puppet_module, module_options|
       begin
+        puppet_module = puppet_module.to_s
```
ekohl, line 174: Can `managed_modules()` return puppet modules as strings? Or if `manage_module` needs strings, should that perform the cast?

raphink: Actually, I don't think modules need to be strings, they can be managed internally as symbols I think.

raphink: Going to restrict the call to `symbolize_keys` to fix that in a cleaner way.

raphink:
@ekohl I've made 3 commits for recursive keys, examples, and managed_modules. Let me know if you prefer to squash everything together.

raphink:
I don't know if it's just me, I find the cucumber tests extremely hard to debug.

ekohl:
> I don't know if it's just me, I find the cucumber tests extremely hard to debug.

That is not just you. It's very annoying you don't get STDOUT and STDERR when something fails.

ekohl:
> @ekohl I've made 3 commits for recursive keys, examples, and managed_modules. Let me know if you prefer to squash everything together.

I don't mind multiple commits and think making it recursive + examples is fine, but the first commit and the 4th should be squashed. Just a personal preference that I want to avoid commit a making a change and commit b undoing that in the same PR, unless it serves a purpose.

Other than that :+1:

raphink:
I have found the issue: it is possible to list modules in `managed_modules.yml` as an array, without options, which fails when parsing with `Util.symbolize_keys()`. I'll fix that.

raphink:
@ekohl 😉

image

raphink:
There, I don't think I can do much cleaner than that without breaking compat!

ekohl:
I've cheated with manually adding things from https://relishapp.com/cucumber/aruba/v/0-11-0/docs/command/all-output-of-commands-which-were-executed and testing for some string that isn't there. https://relishapp.com/cucumber/aruba/v/0-11-0/docs/command/debug-your-command-in-cucumber-test-run suggests calling `binding.pry` in the actual app, but then you have to know where it stops.

Issue #176, opened 2020-06-02
Define PR/MR label behavior


mhutter:
One thing that caught my eye during #175 

* The **GitHab** implementation will ensure the configured labels are added to the PR when the PR is created, but will NOT add any labels if there is already an open PR around.
* The **GitHub** implementation will always update the existing PR - and fail horribly from looking at the code (note this was already broken before the PR)

How should the behavior be?

1. Only ensure the label is set when creating the MR/PR and be done with it
2. Ensure the specified labels are set even if an MR/PR is already present

I can send a PR for the actual changes.

PR #188, opened 2020-08-06
Fix NoMethodError for --pr option (caused by `module_options = nil`)

### LINKED ISSUE #187: GitLab MR: undefined method `[]' for nil:NilClass (NoMethodError)
I'm running a `msync update --pr foo` command to create a MR on GitLab. The command fails in `modulesync`'s `pr` method:

```ruby
Syncing foo/bar-baz
Cloning repository fresh
Cloning from git@gitlab.example.com:foo/bar-baz.git
Creating new branch managed-configuration-update
Error while updating foo/bar-baz
/usr/local/bundle/gems/modulesync-1.3.0/lib/modulesync.rb:186:in `pr': undefined method `[]' for nil:NilClass (NoMethodError)
	from /usr/local/bundle/gems/modulesync-1.3.0/lib/modulesync.rb:137:in `manage_module'
	from /usr/local/bundle/gems/modulesync-1.3.0/lib/modulesync.rb:174:in `block in update'
	from /usr/local/bundle/gems/modulesync-1.3.0/lib/modulesync.rb:172:in `each'
	from /usr/local/bundle/gems/modulesync-1.3.0/lib/modulesync.rb:172:in `update'
	from /usr/local/bundle/gems/modulesync-1.3.0/lib/modulesync/cli.rb:142:in `update'
	from /usr/local/bundle/gems/thor-1.0.1/lib/thor/command.rb:27:in `run'
	from /usr/local/bundle/gems/thor-1.0.1/lib/thor/invocation.rb:127:in `invoke_command'
	from /usr/local/bundle/gems/thor-1.0.1/lib/thor.rb:392:in `dispatch'
	from /usr/local/bundle/gems/thor-1.0.1/lib/thor/base.rb:485:in `start'
	from /usr/local/bundle/gems/modulesync-1.3.0/bin/msync:8:in `<top (required)>'
	from /usr/local/bundle/bin/msync:23:in `load'
	from /usr/local/bundle/bin/msync:23:in `<main>'
```
Related source: [lib/modulesync.rb, line 186](https://github.com/voxpupuli/modulesync/blob/1.3.0/lib/modulesync.rb#L186)

It looks like `module_options` is still unset [when used in the call to `pr`](https://github.com/voxpupuli/modulesync/blob/1.3.0/lib/modulesync.rb#L137) on line 137.

bittner:
Adds `--noop` support for the `--pr` option.

Adds feature tests for both GitHub PRs and GitLab MRs to prove the bug reported with issue #187.

Fixes #187

### DIFF
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -185,13 +185,9 @@ module ModuleSync # rubocop:disable Metrics/ModuleLength
   end
 
   def self.pr(module_options)
-    if !module_options.nil?
-      github_conf = module_options[:github]
-      gitlab_conf = module_options[:gitlab]
-    else
-      github_conf = nil
-      gitlab_conf = nil
-    end
+    module_options ||= {}
+    github_conf = module_options[:github]
+    gitlab_conf = module_options[:gitlab]
 
     if !github_conf.nil?
       base_url = github_conf[:base_url] || ENV.fetch('GITHUB_BASE_URL', 'https://api.github.com')
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -132,6 +132,7 @@ module ModuleSync # rubocop:disable Metrics/ModuleLength
 
     if options[:noop]
       Git.update_noop(git_repo, options)
+      options[:pr] && pr(module_options).manage(namespace, module_name, options)
     elsif !options[:offline]
       pushed = Git.update(git_repo, files_to_manage, options)
       pushed && options[:pr] && pr(module_options).manage(namespace, module_name, options)
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -184,9 +184,14 @@ module ModuleSync # rubocop:disable Metrics/ModuleLength
     exit 1 if errors && options[:fail_on_warnings]
   end
 
-  def self.pr(module_options = {})
-    github_conf = module_options[:github]
-    gitlab_conf = module_options[:gitlab]
+  def self.pr(module_options)
+    if !module_options.nil?
+      github_conf = module_options[:github]
+      gitlab_conf = module_options[:gitlab]
+    else
+      github_conf = nil
+      gitlab_conf = nil
+    end
 
     if !github_conf.nil?
       base_url = github_conf[:base_url] || ENV.fetch('GITHUB_BASE_URL', 'https://api.github.com')
--- a/lib/modulesync/pr/github.rb
+++ b/lib/modulesync/pr/github.rb
@@ -18,18 +18,22 @@ module ModuleSync
         head = "#{namespace}:#{options[:branch]}"
         target_branch = options[:pr_target_branch] || 'master'
 
-        pull_requests = @api.pull_requests(repo_path, :state => 'open', :base => target_branch, :head => head)
-        if pull_requests.empty?
-          pr = @api.create_pull_request(repo_path,
-                                        target_branch,
-                                        options[:branch],
-                                        options[:pr_title],
-                                        options[:message])
-          $stdout.puts \
-            "Submitted PR '#{options[:pr_title]}' to #{repo_path} - merges #{options[:branch]} into #{target_branch}"
+        if options[:noop]
+          puts "Using no-op. Submitted PR '#{options[:pr_title]}' to #{repo_path} - merges #{options[:branch]} into #{target_branch}"
         else
-          # Skip creating the PR if it exists already.
-          $stdout.puts "Skipped! #{pull_requests.length} PRs found for branch #{options[:branch]}"
+          pull_requests = @api.pull_requests(repo_path, :state => 'open', :base => target_branch, :head => head)
+          if pull_requests.empty?
+            pr = @api.create_pull_request(repo_path,
+                                          target_branch,
+                                          options[:branch],
+                                          options[:pr_title],
+                                          options[:message])
+            $stdout.puts \
+              "Submitted PR '#{options[:pr_title]}' to #{repo_path} - merges #{options[:branch]} into #{target_branch}"
+          else
+            # Skip creating the PR if it exists already.
+            $stdout.puts "Skipped! #{pull_requests.length} PRs found for branch #{options[:branch]}"
+          end
         end
 
         # PR labels can either be a list in the YAML file or they can pass in a comma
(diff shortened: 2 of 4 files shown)


lib/modulesync/pr/github.rb
```diff
@@ -18,18 +18,22 @@ def manage(namespace, module_name, options)
...
-        if pull_requests.empty?
-          pr = @api.create_pull_request(repo_path,
-                                        target_branch,
-                                        options[:branch],
-                                        options[:pr_title],
-                                        options[:message])
-          $stdout.puts \
-            "Submitted PR '#{options[:pr_title]}' to #{repo_path} - merges #{options[:branch]} into #{target_branch}"
+        if options[:noop]
+          puts "Using no-op. Submitted PR '#{options[:pr_title]}' to #{repo_path} - merges #{options[:branch]} into #{target_branch}"
```
raphink, line 22: I don't find this really clear. This message seems to say that even though `noop` is used, a PR is being actually submitted anyway…

ekohl: Agreed. Using `would` feels clearer.
```suggestion
          puts "Using no-op. Would submit PR '#{options[:pr_title]}' to #{repo_path} - merges #{options[:branch]} into #{target_branch}"
```

bittner: True. I had this with "would".

The reason why I switched it back to the same wording as the actual, non-noop output is for the tests: It makes it easy to check for the same sentence.

But, really, this is a hack, true. The test should verify what is actually being tested. Currently, this is "slight cheating". I'll fix it.

bittner: Done. Note that the tests are failing _as expected_. It demonstrates that there is an issue and the tests are able to catch it.

bittner:
This should be ready for merging now. Can you take a look again? Thanks!

lib/modulesync.rb
```diff
@@ -183,9 +184,14 @@ def self.update(options)
     exit 1 if errors && options[:fail_on_warnings]
   end
 
-  def self.pr(module_options = {})
-    github_conf = module_options[:github]
-    gitlab_conf = module_options[:gitlab]
+  def self.pr(module_options)
+    if !module_options.nil?
```
ekohl, line 188: Would the code be easier if we used `module_options ||= {}`

lib/modulesync.rb
```diff
@@ -183,9 +184,14 @@ def self.update(options)
     exit 1 if errors && options[:fail_on_warnings]
   end
 
-  def self.pr(module_options = {})
-    github_conf = module_options[:github]
-    gitlab_conf = module_options[:gitlab]
+  def self.pr(module_options)
```
ekohl, line 187: Why is it needed to change the signature?

lib/modulesync.rb
raphink: This is pretty much what I had in the signature before.

As a reminder, the reason for this was to support modules without options (when `managed_modules.yaml` is a list, not a hash).

bittner: The signature alone didn't fix the issue (as demonstrated by the tests I added). The problem is, when you set the method's parameter to `nil` explicitly the default value doesn't help. The change should now lead to the same result, if I see this correctly.

As for the syntactic sugar, I'm not a Ruby person, so feel free to direct me to what is a shorter, equivalent way of writing the code.

lib/modulesync.rb
bittner: The changes to the code makes the default value unnecessary (and potentially misleading). We _could_ leave it in, but it's technically unneeded, I believe.

lib/modulesync.rb
bittner: I applied the syntactic sugar. Looks much nicer now, thanks! :+1:

bittner:
> I wonder if the following pattern would be easier:

I have applied your suggested changes to the GitLab pr module. That makes the code structure flatter, which is better. Nice, thank you!

In the [GitHub pr module](https://github.com/voxpupuli/modulesync/pull/188/files#diff-67ea73e264f66bd81ae8cc64cbbb0d38) the code is different: The code, which sets the labels, needs to run afterwards and has a separate `:noop` clause. It might be a good idea to align the two implementations, e.g. while fixing issue #176, so that code could also benefit from a flatter structure.

bittner:
Any more corrections needed? Otherwise, I'd be happy if we could merge the changes. :crossed_fingers:

raphink:
> I wonder if the following pattern would be easier:
> 
> ```ruby
> do_prep_work
> 
> if options[:noop]
>   $stdout.puts "Using no-op. Would do real work"
>   return
> end
> 
> do_real_work
> ```
> 
> I don't know the answer - there are really good arguments for both.

I was going to say the same actually. I prefer short circuits for this kind of logic.

lib/modulesync/pr/github.rb
```diff
@@ -39,8 +46,12 @@ def manage(namespace, module_name, options)
         # We only assign labels to the PR if we've discovered a list > 1. The labels MUST
         # already exist. We DO NOT create missing labels.
         return if pr_labels.empty?
-        $stdout.puts "Attaching the following labels to PR #{pr['number']}: #{pr_labels.join(', ')}"
-        @api.add_labels_to_an_issue(repo_path, pr['number'], pr_labels)
+        if options[:noop]
+          puts "Using no-op. Would attach the following labels to the PR: #{pr_labels.join(', ')}"
```
raphink, line 50: You could return here as well, to avoid the `else` clause next line.

bittner: I'd love to do that, but this is not easily possible. In the [GitHub pr module](https://github.com/voxpupuli/modulesync/pull/188/files#diff-67ea73e264f66bd81ae8cc64cbbb0d38) the code, which sets the labels, needs to run afterwards and has a separate `:noop` clause. Hence, we can't short-circuit, we need to let the flow hop over the hot code.

Or what would be your concrete suggestion?

ekohl: Somewhat unrelated, but if a PR already existed, `pr` is `nil` in the `else` clause. A fix for that has implications for noop. Either you will `return unless pull_requests.empty?` and that means the noop print at the beginning can also return. The alternative is that you add `pr = pull_requests.first` and the label code should print 'Would ensure labels x,y,z on PR'.

bittner: Right. Turns out there are quite a lot of those "early exit" cases. I think I know addressed them all.

As a result the implementation for GitHub PRs and GitLab MRs has become almost identical. (The API is different for PR/MR labels; GitLab creates them _with_ the MR, GitHub adds them _after_ creation of the PR.)

I hope we're now ready for merging. Can you all take a look again, please?

ekohl: With the current implementation this is unreachable code. On line 25 there's a return after `if options[:noop]`. That's fine, but it means you can drop this line now.

bittner: True! I was totally keen on removing that clause and align the both implementations, and in the end I forgot the one and important thing! :laughing:

Thanks for catching that! :100:

raphink:
Ok for me then

bittner:
Thank you @raphink and @ekohl! :1st_place_medal:

PR #191, opened 2020-08-15
Drop support for Ruby 2.4 and older

bastelfreak:


bastelfreak:
ups, forgot that we've the ruby version in the gemspec as well. updated it.

ekohl:
@raphink any concerns?

raphink:
None whatsoever @ekohl 👍🏼

PR #193, opened 2020-08-18
Drop Ruby 2.0 compatibility

bastelfreak:
Ruby 2.0 is EoL since a long long time and the gem dependencies don't
resolve anymore.


bastelfreak:
should be obsolete with https://github.com/voxpupuli/modulesync/pull/191

PR #194, opened 2020-09-17
Use remote_branch for PRs when specified

raphink:
When `remote_branch` is specified, the PR/MR should be opened against it, not the source branch.

### DIFF
--- a/lib/modulesync/pr/github.rb
+++ b/lib/modulesync/pr/github.rb
@@ -15,13 +15,14 @@ module ModuleSync
 
       def manage(namespace, module_name, options)
         repo_path = File.join(namespace, module_name)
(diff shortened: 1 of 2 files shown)


bastelfreak (approved):
sounds good, but is this a backwards incompatible change?

raphink:
I wouldn't say it's backwards incompatible @bastelfreak, I'd say it's a bugfix, as it makes no sense to ignore `remote_branch` in the PR when it's provided, and the behavior is unchanged when `remote_branch` is not specified.

raphink:
@ekohl any comment on that, or shall I merge?

ekohl (approved):
Makes sense.

Issue #196, opened 2020-10-06
Please adjust the dependency to puppet-blacksmith's current version


dbro86:
Dear developers,
Is it possible to increase the dependency to puppet-blacksmith's current version 6.0.0?

Kind regards
Brody

ekohl:
This repo is the modulesync software itself rather than the config Voxpupuli uses. Perhaps you're looking for https://github.com/voxpupuli/modulesync_config.

bastelfreak:
oh, we haven't updated the dependency in years :( PR is https://github.com/voxpupuli/modulesync/pull/197

PR #197, opened 2020-10-06
Allow newer puppet-blacksmith versions

bastelfreak:
fixes https://github.com/voxpupuli/modulesync/issues/196

### DIFF
--- a/modulesync.gemspec
+++ b/modulesync.gemspec
@@ -26,6 +26,6 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'git', '~>1.3'
   spec.add_runtime_dependency 'gitlab', '~>4.0'
   spec.add_runtime_dependency 'octokit', '~>4.0'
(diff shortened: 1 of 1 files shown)


modulesync.gemspec
```diff
@@ -26,6 +26,6 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'git', '~>1.3'
   spec.add_runtime_dependency 'gitlab', '~>4.0'
   spec.add_runtime_dependency 'octokit', '~>4.0'
-  spec.add_runtime_dependency 'puppet-blacksmith', '~>3.0'
+  spec.add_runtime_dependency 'puppet-blacksmith', '>= 3.0'
```
ekohl, line 29: Should we at least set an upper bound?
```suggestion
  spec.add_runtime_dependency 'puppet-blacksmith', '>= 3.0', '< 7'
```

ekohl (approved):

PR #200, opened 2020-12-16
Fix language-dependent Git output handling

### LINKED ISSUE #85: Language sensitive GIT handling
The line here is language specific, and would break on non-english.

``` ruby
if git_error.message.include? 'nothing to commit, working directory clean'
```

https://github.com/voxpupuli/modulesync/blob/master/lib/modulesync/git.rb#L129

Only help for now is:

```
LANG=C msync update ...
```

neomilium:
When running `msync` on a system configured with a different language than english, some errors catch fail.

This PR fixes #85

### DIFF
--- a/lib/monkey_patches.rb
+++ b/lib/monkey_patches.rb
@@ -38,7 +38,20 @@ module Git
     end
   end
 
+  module LibMonkeyPatch
+    # Monkey patch set_custom_git_env_variables due to our ::Git::GitExecuteError handling.
+    #
+    # We rescue on the GitExecuteError and proceed differently based on the output of git.
+    # This way makes code language-dependent, so here we ensure that Git gem throw git commands with the "C" language
+    def set_custom_git_env_variables
+      super
+      ENV['LANG'] = 'C.UTF-8'
+    end
+  end
+
   class Lib
+    prepend LibMonkeyPatch
+
     # Monkey patch ls_files until https://github.com/ruby-git/ruby-git/pull/320 is resolved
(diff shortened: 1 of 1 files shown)


lib/monkey_patches.rb
```diff
@@ -38,7 +38,20 @@ def process_full_diff
...
   end
 
+  module LibMonkeyPatch
+    # Monkey patch set_custom_git_env_variables due to our ::Git::GitExecuteError handling.
+    #
+    # We rescue on the GitExecuteError and proceed differently based on the output of git.
+    # This way makes code language-dependent, so here we ensure that Git gem throw git commands with the "C" language
+    def set_custom_git_env_variables
+      super
+      ENV['LANG'] = 'C'
```
ekohl, line 48: I'm worried that this will break UTF-8 encoded values. Some distros build `C.UTF-8` but others don't. Any thoughts about this?

neomilium: Git uses `gettext` to translate its messages.
So you could set `LANG=C.UTF-8` to be sure to preserve UTF-8 encoded value, and on distros were `C.UTF-8` is not available it will fallback on untranslated message which will work in our case.

Are you OKay with this assertion?

ekohl (approved):
Looking at the source (https://github.com/ruby-git/ruby-git/blob/f9abb180b84eddbc31ef578d4587b8612455afa3/tests/test_helper.rb#L88-L101), it looks like it restores `ENV` after so this should be safe.

PR #202, opened 2020-12-20
Improve tests capabilities by using local/fake remote repositories

neomilium:
During the writing of #199 , I needed to refactor some code and would like to be bug-free so I relied on tests.

Behavior tests were pretty rich but lack of capabilities to test repository alteration (ie. commits made by `msync`).

I started to work in this when I realized they used online repos where we can not push on then test operations.
Some tests were using a local clone of them to operate effectively on it, but still relies on online dummy repos.

This PR provides a new way to test `msync` operations by faking remote repositories.

It relies on a new class to fake a remote repository of a puppet module.
This class `Faker::PuppetModuleRemoteRepo` (sorry for the long name), provides some features that helps to test `ModuleSync` behavior:
 - Fake a remote repository by creating a bare initialization usable using a local remote url (ie. file://)
 - Create basic operations behind the scene using a temporary repo (e.g. write, add and push a file; read a file from a specified branch)
 - Populate the fake remote repo using a basic metadata.json (that can be useful later to test operations like `bump`)
 - Simulate a read-only repo
 - Change the default branch
 - Count commits between two commit refs
 - Count commits made by an author on all the repo, or a specified branch

Using this, we can now ensure that `msync`, for example, create a commit or not, depending on excepted behavior.
This PR includes these new expectations on existing tests.

As side effect, all tests are now offline which allow to run them faster, specially when connectivity with GitHub is slow.
For the record, ATM, behind a pretty good connection tests passed in 1m11.826s (using current `master` branch), while they now run in 0m42.832s (reported by `cucumber`).

Additionally, this PR replaces `cat` usage by `aruba` file matcher, reduces the verbosity of tests to create a basic setup with one puppet module and finally silent down `cucumber` which promote its online service.

@reviews I tried to split at most this PR with meaningful (I hope so) commit logs; I recommend to review each commit to keep a good track of changes.

Enjoy \o/

PR #204, opened 2020-12-21
Fix exit status code on failures

neomilium:
On top of #202 , this PR fixes exit status code on failures.

Previously, `msync` often returns `0` when errors occurred, this PR sets it to `1` and turn `puts` into `warn` in order to output the error on `stderr` instead of `stdout`.

### DIFF
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -39,10 +39,10 @@ module ModuleSync # rubocop:disable Metrics/ModuleLength
           .collect { |p| p.chomp('.erb') }
           .to_a
     else
-      $stdout.puts "#{local_template_dir} does not exist." \
+      $stderr.puts "#{local_template_dir} does not exist." \
         ' Check that you are working in your module configs directory or' \
         ' that you have passed in the correct directory with -c.'
-      exit
+      exit 1
     end
   end
 
@@ -53,9 +53,9 @@ module ModuleSync # rubocop:disable Metrics/ModuleLength
   def self.managed_modules(config_file, filter, negative_filter)
     managed_modules = Util.parse_config(config_file)
     if managed_modules.empty?
-      $stdout.puts "No modules found in #{config_file}." \
+      $stderr.puts "No modules found in #{config_file}." \
         ' Check that you specified the right :configs directory and :managed_modules_conf file.'
-      exit
+      exit 1
     end
     managed_modules.select! { |m| m =~ Regexp.new(filter) } unless filter.nil?
     managed_modules.reject! { |m| m =~ Regexp.new(negative_filter) } unless negative_filter.nil?
--- a/lib/modulesync/cli.rb
+++ b/lib/modulesync/cli.rb
@@ -1,10 +1,12 @@
 require 'thor'
+
 require 'modulesync'
+require 'modulesync/cli/thor'
 require 'modulesync/constants'
 require 'modulesync/util'
 
 module ModuleSync
-  class CLI
+  module CLI
     def self.defaults
       @defaults ||= Util.symbolize_keys(Util.parse_config(Constants::MODULESYNC_CONF_FILE))
     end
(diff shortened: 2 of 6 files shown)


neomilium:
Mark this as draft waiting for PR #202 to be merged first.

lib/modulesync.rb
```diff
@@ -53,9 +53,9 @@ def self.relative_names(file_list, path)
   def self.managed_modules(config_file, filter, negative_filter)
     managed_modules = Util.parse_config(config_file)
     if managed_modules.empty?
-      $stdout.puts "No modules found in #{config_file}." \
+      warn "No modules found in #{config_file}." \
         ' Check that you specified the right :configs directory and :managed_modules_conf file.'
-      exit
+      exit 1
```
bastelfreak, line 58: what do you think about using a different exit code here? On the long run this might make it easier to automate / parse msync runs? Example from the `puppet agent` man page

```
       •   --detailed-exitcodes: Provide extra information about the run via exit codes; works only if '--test' or '--onetime' is also specified. If enabled, 'puppet agent' uses the following exit codes:

           0: The run succeeded with no changes or failures; the system was already in the desired state.

           1: The run failed, or wasn't attempted due to another run already in progress.

           2: The run succeeded, and some resources were changed.

           4: The run succeeded, and some resources failed.

           6: The run succeeded, and included both changes and failures.
```

lib/modulesync/cli.rb
```diff
@@ -1,10 +1,12 @@
 require 'thor'
+
 require 'modulesync'
+require 'modulesync/cli/thor'
 require 'modulesync/constants'
 require 'modulesync/util'
 
 module ModuleSync
-  class CLI
+  module CLI
```
bastelfreak, line 9: just curious, can you explain why you switched from a class to a module?

lib/modulesync.rb
neomilium: Interesting idea, ATM my goal was to fail when relevant.
AFAICS, this PR already matches with the `puppet agent` exit codes, as it all the fixed codes are set to `1`  and match with a run failed.

lib/modulesync/cli.rb
neomilium: `CLI` is never instantiated and is only here as a namespace, so switching to module is, for me, a good ruby coding practice.

lib/modulesync.rb
```diff
@@ -39,10 +39,10 @@ def self.find_template_files(local_template_dir)
           .collect { |p| p.chomp('.erb') }
           .to_a
     else
-      $stdout.puts "#{local_template_dir} does not exist." \
+      warn "#{local_template_dir} does not exist." \
```
ekohl, line 42: This is [Kernel.warn](https://apidock.com/ruby/Kernel/warn), right? I'm not sure if that's the correct thing to do. Other places in this file use `$stdout.puts` and `$stderr.puts` so I'd prefer to remain consistent. In this case I think it should be `$stderr.puts`. The same below.

The reason is that you can disable all warnings when you run `ruby -W0` which I sometimes do if there's some deprecation warning. Some projects generate them because they're fully Ruby 2.7 clean. That would hide this output.

ekohl (changes requested):
Small nit, otherwise :+1:

bastelfreak (approved):

lib/modulesync.rb
neomilium: My wish was to write clean code and `rubocop` recent version advises to turn `$stderr.puts` int `warn` but you're right: `warn` can be disabled using `ruby -W0`.
So, I will revert the turn from `$stderr.puts` into `warn` as its more an error than a warning.
Thanks for notice.

Source: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/StderrPuts

bastelfreak (approved):

ekohl (approved):

ekohl:
Thanks!

PR #205, opened 2020-12-22
Switch from Travis CI to GitHub Actions

neomilium:
ATM, this PR drops the IRC notification feature.

Tell me if you still want it.


neomilium:
Converting to draft and waiting on PR #202 to be merged, as checks failed for an unknown reason on top of `master` but runs well on top of `improve-tests`.

https://github.com/opus-codium/modulesync/pull/1

neomilium:
Before merging this PR, maintainers need to add secret keys for releasing process (i.e. in `.github/workflows/release.yml`).

Remind: The IRC notification feature is dropped.

Please consider this contribution as a base, not a fully tested code, multiple reviews appreciated.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,26 @@
+on:
```
ekohl, line 1: It looks like you don't have a name for this and the filename isn't great.

ekohl (changes requested):
I pushed the branch to upstream which triggers the tests.

.github/workflows/ci.yml
bastelfreak: @neomilium can you take a look at this?

neomilium: It seems I will not have so much spare time during next weeks. So, feel free to rework my PR if you want to merge it quickly.

neomilium: @ekohl What name you prefer `ci`, `Continuous integration`, `Tests`? Or maybe you have better suggestion. Thx

ekohl: I'm leaning to keeping it short and use `CI`.

neomilium: OK, so `CI` for the name but you point me that `ci.yml` was not a good filename. What did you expect here? ;-)

ekohl: I can't link back to anymore, but if you go to Checks, you see this:

image

Previously that showed the filename instead of `CI`. Looking at it now, Continuous integration would probably also have been fine since there's sufficient space. I hope that helps to understand why I cared about it.

ekohl (approved):
Thanks!

.github/workflows/ci.yml
neomilium: OK, thx for explanations.

PR #208, opened 2021-03-26
Correctly state which config file to update

bittner:
Fixes a minor content mistake in our README:

The `managed_modules.yaml` only lists the repository references. Configuration parameters are set in `modulesync.yml`.


ekohl (approved):

ekohl:
Thanks!

PR #206, opened 2020-12-31
Refactor code for maintainabilty

neomilium:
This PR improves code structure to help owners to maintain the code and contributors to add features.

About the maintainability, this PR attempts to:
 - reduce code repeats (using classes to hold properties)
 - prevent from some name shadowing
 - reduce passed arguments (e.g. `options` were passed to many methods)
 - make `rubocop` happier ;-)

This PR also adds tests for bump, tag and changelog features.

To help contributors, this PR renames some methods with IMHO more explicit names and organizes code into classes.

Even `msync` have been made to manage Puppet modules, there are some users which use it for other purpose (e.g. @raphink wrote https://dev.to/camptocamp-ops/templating-puppet-control-repositories-3pk7) so I started to split specific code into `PuppetModule` class, while generic one is now in `SourceCode` class.

More details in commit logs.

PR #212, opened 2021-04-21
Refactor puppet modules properties

neomilium:
This commit introduces a new class PuppetModule in order to simplify properties computation.

This new class:
 - allows a single `name`, `namespace` and the working directory computation
 - introduces the `given_name` property, which is the name the user puts in
 its configuration file
 - adds the capability a have a longer `namespace` (e.g. puppet/modules)

This commit also exposes the options through ModuleSync.options instead
of passing it to almost any methods.

As `name` and `namespace` can be confusing in code, so they have been
renamed to `repository_name` and `repository_namespace`.

This commit only introduces minor changes on the output of `msync`: some
single quotes have been added as delimiters for relevant dynamic part of
the messages; and the use of `given_name` (i.e. user-provided name) when
relevant (e.g. "Syncing 'voxpupuli/puppet-module'").


lib/modulesync.rb
```diff
@@ -21,12 +24,16 @@ def self.config_defaults
...
+
   def self.local_file(config_path, file)
     File.join(config_path, MODULE_FILES_DIR, file)
   end
 
-  def self.module_file(project_root, namespace, puppet_module, *parts)
-    File.join(project_root, namespace, puppet_module, *parts)
+  def self.module_file(puppet_module, *parts)
+    File.join(puppet_module.working_directory, *parts)
   end
```
ekohl, line 37: Should this method be one in the `PuppetModule` as well? What triggered me was `module_file(puppet_module, MODULE_CONF_FILE)` which can probably be `puppet_module.path(MODULE_CONF_FILE)` or similar.

lib/modulesync.rb
```diff
@@ -104,38 +110,31 @@ def self.manage_file(filename, settings, options)
...
                             module_configs,
-                            :puppet_module => module_name,
+                            :puppet_module => puppet_module.repository_name,
                             :git_base => options[:git_base],
-                            :namespace => namespace)
+                            :namespace => puppet_module.repository_namespace)
+
     settings.unmanaged_files(module_files).each do |filename|
-      $stdout.puts "Not managing #{filename} in #{module_name}"
+      $stdout.puts "Not managing '#{filename}' in '#{puppet_module.given_name}'"
```
ekohl, line 126: Should the module have a `.to_s` which is `given_name`?

lib/modulesync/git.rb
```diff
@@ -45,23 +45,22 @@ def self.switch_branch(repo, branch)
...
 
-    def self.pull(git_base, name, branch, project_root, opts)
-      puts "Syncing #{name}"
-      Dir.mkdir(project_root) unless Dir.exist?(project_root)
+    def self.pull(puppet_module, branch)
+      puts "Syncing '#{puppet_module.given_name}'"
 
       # Repo needs to be cloned in the cwd
-      if !Dir.exist?("#{project_root}/#{name}") || !Dir.exist?("#{project_root}/#{name}/.git")
+      if !Dir.exist?("#{puppet_module.working_directory}/.git")
```
ekohl, line 52: While badly named, it looks like this should be using `module_file()` as well. Not strictly needed now, but a good future improvement.

lib/modulesync/puppet_module.rb
```diff
@@ -0,0 +1,49 @@
...
+    def repository_path
+      @repository_path ||= "#{repository_namespace}/#{repository_name}"
+    end
+
+    def repository_remote
+      @repository_remote ||= @options[:remote] || _repository_remote
+    end
+
+    def working_directory
+      @working_directory ||= "#{ModuleSync.options[:project_root]}/#{repository_path}"
```
ekohl, line 39:
```suggestion
      @working_directory ||= File.join(ModuleSync.options[:project_root], repository_path)
```

ekohl (commented):
Some minor things inline but I like this abstraction. Cleans up a lot.

Of course some bikeshedding: should it be `PuppetModule`? I know it's used outside of Puppet and nothing inherently ties modulesync to Puppet modules. Perhaps `ManagedModule` is a better name and in line with `managed_modules.yml`.

lib/modulesync.rb
neomilium: You are right, its already planned in next PR. This one is a part of #206

lib/modulesync.rb
neomilium: ATM, not sure it's a good idea: `rubocop` will warn about usage of `to_s` but only mentionning `puppet_module` in string expansion can be confusing.
When refactoring will be done, I can apply your suggestion if you thing its relevant.

lib/modulesync/git.rb
neomilium: OK for futurs improvement, I'll write it down in my TODO file.

lib/modulesync/puppet_module.rb
neomilium: Sorry, I'm a UNIX-only guy ;-)
But `modulesync` seems to be too... isn't it?

Do you really want this?

ekohl: While I'm also UNIX-only, what is the downside of using `File.join`? To me it also signals that first of all, it's a path. I also think the various parts (directories + file) are easier to read than a long string. To me that readability alone is why I always use it, even if I never intend to deploy on Windows.

neomilium: Good point, I followed your suggestion. Done.

ekohl:
> Of course some bikeshedding: should it be `PuppetModule`? I know it's used outside of Puppet and nothing inherently ties modulesync to Puppet modules. Perhaps `ManagedModule` is a better name and in line with `managed_modules.yml`.

Any thoughts on this part?

neomilium:
> Some minor things inline but I like this abstraction. Cleans up a lot.
> 
> Of course some bikeshedding: should it be `PuppetModule`? I know it's used outside of Puppet and nothing inherently ties modulesync to Puppet modules. Perhaps `ManagedModule` is a better name and in line with `managed_modules.yml`.

Yes, I should be `PuppetModule` right now there are some specific code to manage Puppet modules. My goal with #206 is to ease maintenance and to make code more generic... but before this step I'm trying to have a more readable and clean code.
So, for sake of consistency, I started with the right name (IMHO) and will try to slit specific code later.

About `managed_modules.yml` filename, I would like to discuss about it too... but first, I'm focused on improving the code quality/maintainability.

spec/unit/modulesync_spec.rb
```diff
@@ -1,11 +1,13 @@
 require 'spec_helper'
 
+require 'modulesync/puppet_module'
```
ekohl, line 3: Why is this line needed? I'd have expected it to be there since `ModuleSync` is already required and that loads it.

features/cli.feature
```diff
@@ -35,6 +35,9 @@ Feature: CLI
       """
     And a git_base option appended to "modulesync.yml" for local tests
     And a directory named "moduleroot"
-    When I run `msync update --noop --namespace fakenamespace`
+    When I run `msync update --noop --namespace fakenamespace --branch command-line-branch`
```
ekohl, line 38: Kind of feels like an unrelated change. Why is it needed here?

ekohl (approved):
Looks like a good refactor. 2 more small items, but otherwise :+1:

features/cli.feature
neomilium: This test checks that `modulesync` CLI arguments supersede the options specified in configuration file. The way to do this is to check the CLI output... but the output changes (like in this commit), we need to change this test and can be disturbing.

So, to match with the goal of this test, I change it by now checking the branch as the text `Syncing xxx/yyy` now uses the `given_name` instead of `#{namespace}/#{name}`.

ekohl: Personally I would submit it as a separate PR. Not going to push for it now and I'm ok merging this, but small focused PRs tend to be reviewed & merged quicker.

spec/unit/modulesync_spec.rb
neomilium: You're right, its not needed.

neomilium: Done.

features/cli.feature
neomilium: I agree.

ekohl (approved):
Thanks! Let's get this in to allow more.

neomilium:
Ready to merge?

ekohl:
I shouldn't forget to press the big green button. Kind of important ;)

PR #214, opened 2021-04-21
Refactor repository related code

neomilium:
All changes are detailed in commit logs.

Please note that the short log entries of these commits are lack of elegance, so I can merge them or if you have suggestions, I can reword them.

Nevertheless, this PR refactors the formerly `ModuleSync::Git` module in order to clean code and structure it.
This PR is one more part of #206


ekohl (approved):
I like this. It's a common pattern I've seen in multiple places where a module was used instead of a class. When (almost) every method static method has the same argument, you should have used a class instead. Thanks for cleaning this up!

PR #215, opened 2021-04-22
Split generic and specific code

neomilium:
This PR offers the first step to turn `msync` into a more generic tool by splitting `PuppetModule` class into `SourceCode` and `PuppetModule` classes.

PS: This PR is a part of #206.


lib/modulesync/source_code.rb
```diff
@@ -0,0 +1,53 @@
...
+module ModuleSync
+  # Provide methods to retrieve source code attributes
+  class SourceCode
+    attr_reader :given_name
+    attr_reader :options
+    attr_reader :repository
+
+    def initialize(given_name, options)
+      options ||= {}
+      @options = Util.symbolize_keys(options)
```
ekohl, line 14:
```suggestion
      @options = Util.symbolize_keys(options || {})
```

lib/modulesync/source_code.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      @given_name = given_name
+
+      if given_name.include?('/')
+        @repository_name = given_name.split('/').last
+        @repository_namespace = given_name.split('/')[0...-1].join('/')
+      end
+
+      @repository = Repository.new directory: working_directory, remote: repository_remote
```
ekohl, line 23: Should this be lazy initialized?

lib/modulesync/puppet_module.rb
```diff
@@ -1,49 +1,37 @@
...
-    end
-
-    def repository_remote
-      @repository_remote ||= @options[:remote] || _repository_remote
-    end
-
-    def working_directory
-      @working_directory ||= File.join(ModuleSync.options[:project_root], repository_path)
+module ModuleSync
+  # Provide methods to manipule puppet module code
```
ekohl, line 6:
```suggestion
  # Provide methods to manipulate puppet module code
```

lib/modulesync/puppet_module.rb
```diff
@@ -1,49 +1,37 @@
...
+  class PuppetModule < SourceCode
+    def update_changelog(version, message)
+      changelog = "#{working_directory}/CHANGELOG.md"
```
ekohl, line 9: Perhaps use `File.join` while you're at it?
```suggestion
      changelog = File.join(working_directory, 'CHANGELOG.md')
```

lib/modulesync/puppet_module.rb
```diff
@@ -1,49 +1,37 @@
...
+      end
     end
 
-    private
-
-    def _repository_remote
-      git_base = ModuleSync.options[:git_base]
-      git_base.start_with?('file://') ? "#{git_base}#{repository_path}" : "#{git_base}#{repository_path}.git"
+    def bump(message, changelog = false)
+      m = Blacksmith::Modulefile.new("#{working_directory}/metadata.json")
```
ekohl, line 27: Perhaps use `File.join while you're at it?
```suggestion
      m = Blacksmith::Modulefile.new(File.join(working_directory, 'metadata.json'))
```

lib/modulesync/puppet_module.rb
```diff
@@ -1,49 +1,37 @@
...
+      new = m.bump!
+      puts "Bumped to version #{new}"
+      repository.git.add('metadata.json')
+      update_changelog(new, message) if changelog
+      repository.git.commit("Release version #{new}")
+      repository.git.push
```
ekohl, line 33: A bump method that also commits and pushes feels surprising to me.

Maybe you want to tackle this alter, but IMHO there should be one push per repo: it should push both the branch and tag. This makes it more atomic.

ekohl (commented):
Overall this makes sense. I like how you're starting to separate the workflow from low level implementation.

lib/modulesync/puppet_module.rb
neomilium: I agree. My goal during the refactoring process is to introduce only minor changes and fixes bugs, keeping the current behavior.

From my point of view, whole `PuppetModule` class should be deleted from `msync`: most of us uses a local (ie. contained in each puppet module) `rake` task to bump its version, and for changelog more and more use an algorithm to generate it...
What do you think about discuss this point in a dedicated issue?

lib/modulesync/source_code.rb
neomilium: Done.

lib/modulesync/source_code.rb
neomilium: It can be, you're right.
Done.

lib/modulesync/puppet_module.rb
neomilium: Thanks.

lib/modulesync/puppet_module.rb
neomilium: Like you mentioned previously, I turned `ModuleSync::module_file` into `SourceCode#path`, then use it here.
Thanks.

lib/modulesync/puppet_module.rb
neomilium: Same as previous comment.
Done.

lib/modulesync/puppet_module.rb
neomilium: I opened an issue about this. #216

neomilium:
@ekohl This PR is ready to merge IMHO, could you have a look? Thanks

lib/modulesync.rb
```diff
@@ -88,7 +83,7 @@ def self.manage_file(puppet_module, filename, settings, options)
     namespace = settings.additional_settings[:namespace]
     module_name = settings.additional_settings[:puppet_module]
     configs = settings.build_file_configs(filename)
-    target_file = module_file(puppet_module, filename)
+    target_file = puppet_module.path filename
```
ekohl, line 86: I prefer using braces to be explicit and I have the impression that most of the rest of the codebase does too.
```suggestion
    target_file = puppet_module.path(filename)
```

ekohl (approved):
Small nit, but otherwise :+1:

lib/modulesync.rb
neomilium: Done. Thx

neomilium:
@ekohl Thats the moment to push the big green button ;-)

ekohl (approved):
I like I don't have to highlight every case of a pattern. You understand that. I've seen it different in some reviews :)

PR #217, opened 2021-04-23
Rework exception handling

neomilium:
This commit gives the user a better output message:
- it uses the exception message if available or fallback on the "Error during update" message
- it prefixes the message by the current module name

This commits introduces a ModuleSync::Error that will be used to raise
exceptions, with a message outputed on stderr, but without a full
backtrace, useless for user.

It keeps the current behavior when error (ie. StandardError) comes from
templates (ie. ERB) rendering as user may need to see why the rendering
failed.

Git::GitExecuteError backtraces are now hidden too: user needs
to understand what happends, with the error message, but not where error
occured.

### DIFF
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -12,6 +12,8 @@ require 'modulesync/util'
 require 'monkey_patches'
 
 module ModuleSync # rubocop:disable Metrics/ModuleLength
+  class Error < StandardError; end
+
   include Constants
 
   def self.config_defaults
@@ -98,8 +100,8 @@ module ModuleSync # rubocop:disable Metrics/ModuleLength
         }
         template = Renderer.render(erb, configs, metadata)
         Renderer.sync(template, target_file)
-      rescue # rubocop:disable Lint/RescueWithoutErrorClass
-        $stderr.puts "Error while rendering #{filename}"
+      rescue StandardError => e
+        $stderr.puts "#{puppet_module.given_name}: Error while rendering file: '#{filename}'"
         raise
       end
     end
@@ -173,8 +175,13 @@ module ModuleSync # rubocop:disable Metrics/ModuleLength
     managed_modules.each do |puppet_module|
       begin
         manage_module(puppet_module, module_files, defaults)
-      rescue # rubocop:disable Lint/RescueWithoutErrorClass
-        warn "Error while updating '#{puppet_module.given_name}'"
+      rescue ModuleSync::Error, Git::GitExecuteError => e
+        message = e.message || "Error during '#{options[:command]}'"
+        $stderr.puts "#{puppet_module.given_name}: #{message}"
+        exit 1 unless options[:skip_broken]
+        errors = true
+        $stdout.puts "Skipping '#{puppet_module.given_name}' as update process failed"
+      rescue StandardError => e
         raise unless options[:skip_broken]
         errors = true
         $stdout.puts "Skipping '#{puppet_module.given_name}' as update process failed"
--- a/lib/modulesync/repository.rb
+++ b/lib/modulesync/repository.rb
@@ -120,13 +120,10 @@ module ModuleSync
           repo.push('origin', branch, opts_push)
         end
       rescue Git::GitExecuteError => e
-        if e.message.match?(/working (directory|tree) clean/)
-          puts "There were no changes in '#{@directory}'. Not committing."
-          return false
-        else
-          puts e
-          raise
-        end
+        raise unless e.message.match?(/working (directory|tree) clean/)
+
+        puts "There were no changes in '#{@directory}'. Not committing."
+        return false
       end
 
       true


lib/modulesync.rb
```diff
@@ -173,8 +175,14 @@ def self.update(cli_options)
     managed_modules.each do |puppet_module|
       begin
         manage_module(puppet_module, module_files, defaults)
-      rescue # rubocop:disable Lint/RescueWithoutErrorClass
-        warn "Error while updating '#{puppet_module.given_name}'"
+      rescue ModuleSync::Error, Git::GitExecuteError => e
+        message = e.message || "Error during '#{options[:command]}'"
+        message = "#{puppet_module.given_name}: #{message}"
+        $stderr.puts message
```
ekohl, line 181: This may be my taste, but I dislike assigning variables again. Perhaps this is just as good?
```suggestion
        $stderr.puts "#{puppet_module.given_name}: #{message}"
```

lib/modulesync/repository.rb
```diff
@@ -120,13 +120,10 @@ def submit_changes(files, options)
...
         end
       rescue Git::GitExecuteError => e
-        if e.message.match?(/working (directory|tree) clean/)
-          puts "There were no changes in '#{@directory}'. Not committing."
-          return false
-        else
-          puts e
-          raise
-        end
+        raise unless e.message.match?(/working (directory|tree) clean/)
```
ekohl, line 123: I agree this rescue block is a bit weird. Does it make sense to go a step further and prior to committing check if the working tree is dirty? I think this would work
```ruby
unless repo.status.added.any?
  puts "There were no changes in '#{@directory}'. Not committing."
  return false
end

File: lib/modulesync.rb (hunk 1 above)
Comment (neomilium) [reply lib/modulesync.rb:181]: LGTM!

File: lib/modulesync/repository.rb (hunk 2 above)
Comment (neomilium) [reply lib/modulesync/repository.rb:123]: IMHO, this is the whole method `#submit_changes` that should be reworked.
It seems to do many things and some looks useless or weird to me (some examples):
* why amending a commit in this `msync` context?
* should `force` option be the default in `msync` context?
* the `pre_commit_hook` feature is too specific (as said in #127)
* the `remote_branch` handling seems overkill
* finally, the git exception handling...

To address these topics, I think its safer to wrote tests first because ATM, there are no behavior tests we can rely on to improve these points.

So if you are OK, we keep this code which work as-is and we can rework it latter.

File: lib/modulesync.rb (hunk 1 above)
Comment (neomilium) [reply lib/modulesync.rb]: Done.

File: lib/modulesync/repository.rb (hunk 2 above)
Comment (neomilium) [reply lib/modulesync/repository.rb:123]: @ekohl If its OK for you, its ready to be merged.

Comment (ekohl) [reply lib/modulesync/repository.rb:123]: Yes, that feels good. I'm not sure if my code actually **is** the correct solution while yours is "just" moving code around. Generally that's safer. Consider my comment input on a future refactor.

File: lib/modulesync.rb [hunk 3]
```diff
@@ -173,8 +175,13 @@ def self.update(cli_options)
     managed_modules.each do |puppet_module|
       begin
         manage_module(puppet_module, module_files, defaults)
-      rescue # rubocop:disable Lint/RescueWithoutErrorClass
-        warn "Error while updating '#{puppet_module.given_name}'"
+      rescue ModuleSync::Error, Git::GitExecuteError => e
+        message = e.message || "Error during '#{options[:command]}'"
+        $stderr.puts "#{puppet_module.given_name}: #{message}"
+        exit 1 unless options[:skip_broken]
+        errors = true
```
ekohl, line 182: Should this instead be:
```suggestion
        errors = true
        break unless options[:skip_broken]
```

That way the error handling at the end of the method is called (and it respects `options[:fail_on_warnings]`).

lib/modulesync.rb
```diff
@@ -173,8 +175,13 @@ def self.update(cli_options)
     managed_modules.each do |puppet_module|
       begin
         manage_module(puppet_module, module_files, defaults)
-      rescue # rubocop:disable Lint/RescueWithoutErrorClass
-        warn "Error while updating '#{puppet_module.given_name}'"
+      rescue ModuleSync::Error, Git::GitExecuteError => e
+        message = e.message || "Error during '#{options[:command]}'"
```
ekohl, line 179: You can consider wrapping `Git::GitExecuteError` in `ModuleSync::GitError` (which inherits from `ModuleSync::Error` inside `manage_module`.

lib/modulesync.rb
neomilium: No, IMHO, it should not be like this... The trouble comes from distinguish errors and warnings.

Again, this part could be reworked better if we had more behavior tests.

lib/modulesync.rb
neomilium: I cant see the benefit, instead, if we want to improve the exception usage, I think we should catch them at lower level (`repository.rb`) and throw an `ModuleSync::Repository::Error` error or something like this which is more generic but not covered by the improvements I made here.

ekohl: Well, that's sort of what I meant. `ModuleSync::Repository::Error` sounds even better. It would properly encapsulate the lower level operations. Consider it for the future.

neomilium: OK. I wrote it in my todo file (so many thing I want to contribute before...)

neomilium:
@ekohl Big green button?

lib/modulesync.rb
neomilium: From my point of view, we should change our approach about this:
 * `skip_broken` is not relevant option as, for example, an "update" that does not contain modification is consider as broken
 * `skip_broken` sounds me to be the default so turning option from `skip_broken` to `fail_fast` makes me sense to me
 * `fail_on_warnings` could be an option if we choose to introduction a distinguishment between real error (e.g. unable to process a template) and a warning (e.g. no files are updated after an `update` run)

neomilium: As it introduces a big breaking change, I prefer to finalize current work, then discuss about it if needed, then purpose a patch.

neomilium: I opened a new issue about this point: #218

ekohl (approved):

PR #221, opened 2021-05-03
Deal with nil branch

ekohl:
default_branch doesn't guarantee a result. This checks for the nil value and raises an exception.

This is probably not a good way to deal with it, but it's a lot clearer for the user. I happened to run into this because I didn't have a HEAD for the remote `origin` so `rm .git/refs/remotes/origin/HEAD` is the reproducer.

### DIFF
--- a/lib/modulesync/repository.rb
+++ b/lib/modulesync/repository.rb
@@ -39,6 +39,9 @@ module ModuleSync
     def switch_branch(branch)
       unless branch
         branch = default_branch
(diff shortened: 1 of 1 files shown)


bastelfreak:
We can close this since https://github.com/voxpupuli/modulesync/pull/233 got merged?

alexjfisher:
> We can close this since #233 got merged?

Don't think so.

PR #225, opened 2021-06-15
release 2.1.1 /  Create ~/.gem before writing the token to it

bastelfreak:


ekohl (changes requested):
I'd prefer to try `mkdir -p ~/.gem` before using the file.

Also, please align on `~/.gem` rather than using `/home/runner` in the `chmod` command.

ekohl (approved):

PR #226, opened 2021-06-24
Allow latest aruba dependency

bastelfreak:


modulesync.gemspec
```diff
@@ -17,7 +17,7 @@ Gem::Specification.new do |spec|
   spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
   spec.require_paths = ['lib']
 
-  spec.add_development_dependency 'aruba', '~> 0.14'
+  spec.add_development_dependency 'aruba', '>= 0.14'
```
ekohl, line 20: Do we still want to set an upper bound of `< 2` to prevent accepting a new major version accidentally?
```suggestion
  spec.add_development_dependency 'aruba', '>= 0.14', '< 2'
```

ekohl (commented):
Overall :+1: for compat with latest versions

PR #227, opened 2021-06-24
dont install octokit via Gemfile

bastelfreak:
this PR also contains https://github.com/voxpupuli/modulesync/pull/226

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,6 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 gemspec
 
 gem 'cucumber'
-gem 'octokit', '~> 4.9'
 
 group :release do
   gem 'github_changelog_generator', :require => false


ekohl (commented):
Why is 4.9 in Gemfile while the gemspec has 4.0? It's all a bit inconsistent.

bastelfreak:
I think there was a lot of copy and paste ongoing in the past. If you like I can add `'~> 4.9'` to the gemspec file.

Gemfile
```diff
@@ -3,7 +3,6 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 gemspec
 
 gem 'cucumber'
```
ekohl, line 5: Just wondering: shouldn't this be in the gemspec as a dev dependency?

ekohl (approved):
I'm fine not updating it in the gemspec. It's not bound to it and it's probably OK. Technically we could verify what really is the minimum version, but I don't see anyone doing so.

PR #231, opened 2021-07-22
Fix support of repos without a 'master' branch

smortex:
When the target branch does not exist either remotely or locally, the
code checkouts the 'master' branch before branching.  Recently, a number
of projects have switched from naming their main branch 'master' to
'main', and these projects may not have a 'master' branch anymore.  In
such a situation, modulesync failed.

When we are in such a situation, check for a 'main' branch, and if it
exist use this one; otherwise fallback to the previous default of
'master'.

### DIFF
--- a/lib/modulesync/repository.rb
+++ b/lib/modulesync/repository.rb
@@ -51,7 +51,11 @@ module ModuleSync
         repo.checkout("origin/#{branch}")
         repo.branch(branch).checkout
       else
-        repo.checkout('origin/master')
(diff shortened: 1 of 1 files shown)


alexjfisher:
or https://github.com/voxpupuli/modulesync/pull/233 ??

(though your commit message is *much* better than mine!)

smortex:
I like the way #233 has no conditionals nor hardcoded values.  Seems more future-proof :wink:

PR #233, opened 2021-07-22
Checkout default_branch and not hardcoded `master`

alexjfisher:


smortex (approved):

ekohl (commented):
I like this, but I have read that sometimes the `remotes/origin/HEAD` ref isn't fetched. I wonder if `default_branch` should be made more reliable to deal with it. On the other hand, I'd be ok with merging this.

What would it take for this to get out of Draft?

lib/modulesync/repository.rb
```diff
@@ -51,17 +51,21 @@ def switch_branch(branch)
...
         repo.branch(branch).checkout
       end
     end
 
     def prepare_workspace(branch)
       # Repo needs to be cloned in the cwd
       if !Dir.exist?("#{@directory}/.git")
-        puts 'Cloning repository fresh'
-        puts "Cloning from '#{@remote}'"
+        puts "Cloning repository fresh from '#{@remote}'"
```
alexjfisher, line 68: Rubocop was annoyed the class was 1 too many lines, so I 'stole' back one here...

bastelfreak (approved):

ekohl (approved):

PR #244, opened 2021-10-18
New CLI commands

neomilium:
Main goal is to be able to execute arbitrary commands on each repositories.

This PR contains a draft for new commands : `reset`, `execute` and `push`

* `clone`: clone repo if not already done
* `execute`: run arbitrary command
* `reset`: hard-reset repositories
* `push`: push commits

Any help to define expected behaviors is welcome.


Gemfile.lock
```diff
@@ -0,0 +1,263 @@
+PATH
```
ekohl, line 1: Generally we don't store a lock file. Could you drop this?

ekohl (commented):
I would still love to see the `execute` command. It would still help me a lot.

neomilium:
Code rebased, contributions are welcome, dev is stalled on this ;-)

lib/modulesync.rb
```diff
@@ -177,4 +178,83 @@ def self.update(cli_options)
...
+
+      command_args = cli_options[:command_args]
+      local_script = File.expand_path command_args[0]
+      command_args[0] = local_script if File.exist?(local_script)
+
+      # Remove bundler-related env vars to allow the subprocess to run `bundle`
+      command_env = ENV.reject { |k, _v| k.match?(/(^BUNDLE|^SOURCE_DATE_EPOCH$|^GEM_|RUBY)/) }
+
+      FileUtils.chdir(puppet_module.working_directory) do
+        result = system command_env, *command_args, unsetenv_others: true
```
ekohl, line 208: Rather than using `chdir`, you can pass `chdir` to system
```suggestion
      result = system(command_env, *command_args, unsetenv_others: true, chdir: puppet_module.working_directory)
```

Gemfile.lock
neomilium: Done.

features/execute.feature
```diff
@@ -0,0 +1,15 @@
...
+  Scenario: Cloning sourcecodes before running command when modules/ dir is empty
+    Given a basic setup with a puppet module "puppet-test" from "awesome"
+    Then the file "modules/awesome/puppet-test/metadata.json" should not exist
+    When I successfully run `msync exec --verbose /bin/true`
+    Then the output should contain "Cloning from 'file://"
+    And the file "modules/awesome/puppet-test/metadata.json" should exist
+
+  @no-clobber
+  Scenario: No clones before running command when sourcecode have already been cloned
+    Given the file "modules/awesome/puppet-test/metadata.json" should exist
```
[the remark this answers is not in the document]
neomilium: Done.

bastelfreak:
@neomilium do you still consider this a draft? can you please take a look at the rubocop issues?

neomilium:
@bastelfreak Yes, its still a draft, it requires to be used and feedbacks are needed.

Rubocop issues should be fixed.

neomilium:
One command is missing to the set to be really usable: `pull-request` or `merge-request`…

lib/modulesync.rb
```diff
@@ -177,4 +178,83 @@ def self.update(cli_options)
...
+        unless result
+          message = "Command execution failed ('#{@options[:command_args].join ' '}': #{$CHILD_STATUS})"
+          raise Thor::Error, message if @options[:fail_fast]
+
+          errors.merge(
```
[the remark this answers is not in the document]
neomilium: Fixed

neomilium:
@bastelfreak @ekohl Enjoy!
Please note I did not update the README... IMHO, whole file could be rewritten.

lib/modulesync/repository.rb
```diff
@@ -63,22 +63,24 @@ def switch_branch(branch)
...
+    def clone
+      puts "Cloning from '#{@remote}'"
+      @git = Git.clone(@remote, @directory)
+    end
+
     def prepare_workspace(branch)
-      # Repo already cloned, check out master and override local changes
-      if Dir.exist? File.join(@directory, '.git')
-        # Some versions of git can't properly handle managing a repo from outside the repo directory
-        Dir.chdir(@directory) do
```
ekohl, line 70: The chdir is no longer needed? I tried looking at the ruby-git code but never looked deep enough to figure out if the bug had been resolved.

lib/modulesync/repository.rb
```diff
@@ -72,14 +72,16 @@ def clone
...
-    def prepare_workspace(branch)
+    def prepare_workspace(branch:, operate_offline:)
       if cloned?
         puts "Overriding any local changes to repository in '#{@directory}'"
-        git.fetch 'origin', prune: true
+        git.fetch 'origin', prune: true unless operate_offline
         git.reset_hard
         switch(branch: branch)
-        git.pull('origin', branch) if remote_branch_exists?(branch)
+        git.pull('origin', branch) if remote_branch_exists?(branch) && !operate_offline
```
ekohl, line 81: I'd swap the order and first check `!operate_offline`. I'm not sure if `remote_branch_exists?` reaches out to the network, but it's certainly cheaper to check a boolean first.

lib/modulesync.rb
```diff
@@ -108,7 +108,7 @@ def self.manage_file(puppet_module, filename, settings, options)
 
   def self.manage_module(puppet_module, module_files, defaults)
     puts "Syncing '#{puppet_module.given_name}'"
-    puppet_module.repository.prepare_workspace(options[:branch]) unless options[:offline]
+    puppet_module.repository.prepare_workspace(branch: options[:branch], operate_offline: options[:offline])
```
ekohl, line 112: This is changing behavior. I always used `msync update --offline` as "do not do any git actions", or `update --no-git`. It may not be entirely right, but it's a behavioral change.

lib/modulesync.rb
```diff
@@ -177,4 +178,29 @@ def self.update(cli_options)
...
+      $stdout.puts "#{puppet_module.given_name}:"
+
+      puppet_module.repository.clone unless puppet_module.repository.cloned?
+      puppet_module.repository.switch branch: @options[:branch]
+
+      command_args = cli_options[:command_args]
+      local_script = File.expand_path command_args[0]
+      command_args[0] = local_script if File.exist?(local_script)
+
+      FileUtils.chdir(puppet_module.working_directory) do
```
ekohl: What's the benefit of `FileUtils.chdir` over `Dir.chdir`?

lib/modulesync.rb
```diff
@@ -177,4 +178,29 @@ def self.update(cli_options)
...
+
+    managed_modules.each do |puppet_module|
+      $stdout.puts "#{puppet_module.given_name}:"
+
+      puppet_module.repository.clone unless puppet_module.repository.cloned?
+      puppet_module.repository.switch branch: @options[:branch]
+
+      command_args = cli_options[:command_args]
+      local_script = File.expand_path command_args[0]
+      command_args[0] = local_script if File.exist?(local_script)
```
ekohl, line 202: I'm not a fan of this. I'd prefer to be more explicit in the behavior. If it's a path (on unix: it starts with `/`, `./` or `../`), only then would I treat it as a script. And `system()` can already do that. So personally I'd leave this out.

lib/modulesync/cli.rb
```diff
@@ -169,6 +169,32 @@ def execute(*command_args)
         ModuleSync.execute(config)
       end
 
+      desc 'reset', 'Reset all repositories'
```
ekohl, line 173: What does reset mean? Looking at the code I can figure it out but I think you can be a bit more verbose.

lib/modulesync/repository.rb
```diff
@@ -163,6 +163,17 @@ def submit_changes(files, options)
       true
     end
 
+    def push(branch:, remote_branch:)
+      raise ModuleSync::Error, 'Repository must be locally available before trying to push' unless cloned?
+
+      remote = 'origin'
```
ekohl, line 169: I wish we had a constant for this, but I don't expect you to do that in this PR. However, perhaps in this PR it can be, at least in the private method a parameter with a default (so `push(..., remote_name: 'origin')` or similar) so it will be easier in the future?

lib/modulesync/cli.rb
```diff
@@ -7,6 +7,14 @@
 
 module ModuleSync
   module CLI
+    def self.prepare_options(cli_options, **more_options)
+      options = CLI.defaults
+      options.merge! Util.symbolize_keys cli_options
```
ekohl, line 12: I feel like braces here would be helpful. In general I'm not fan of skipping them except when you do DSL style things.

ekohl (commented):
My biggest concern is the change in `msync update --offline` behavior, the other things are less important.

lib/modulesync/repository.rb
neomilium: Never faced myself to the bug, but this line is now old, and AFAICR there is another line that do it with `chdir`

lib/modulesync.rb
neomilium: Hum, as the _local script_ is relative to the top level (ie. `modulesync_config`) and no repository one, the `system()` command will fail. That's why I expanded the path before using it.

lib/modulesync/repository.rb
neomilium: And tests are two…

ekohl: I'd be ok with that.

lib/modulesync.rb
neomilium: OK. Based on existing tests, I understood `offline` as a flag to operate offline, not as operate without git operations.

If I understand correctly you use `msync update --offline` to only apply templates without any modifications, could you confirm or precise the goal you want or reach using this command line?

lib/modulesync.rb
neomilium: In our context, none.
I toggle this conversation as this is a commit on top of this which remove this line.

lib/modulesync.rb
ekohl: It is indeed. Essentially, I'd like a way to run the templates on my checkout without involving git. This is because I already prepared the whole repo, may have done some manual modifications. For example, I could have gone in and used `git add` and then use `git diff --cached` to see the exact changes. Or I did some preparation PRs to a branch. Essentially make it just a templating tool.

neomilium: So maybe a `msync templates apply` command could be a more explicit command to reach your goal?

lib/modulesync/cli.rb
neomilium: Done: I explained it in long description

lib/modulesync/repository.rb
neomilium: Done.

lib/modulesync/cli.rb
neomilium: Done.

lib/modulesync.rb
ekohl: Probably, but that would be a major version bump. I wouldn't mind that, but I'd hate to lose the functionality now.

lib/modulesync/repository.rb
neomilium: Good catch, thanks. Suggestion applied.

lib/modulesync.rb
ekohl: Hmm, it does make sense but I can also see how it would be confusing. I'm not sure how to best deal with this. Perhaps document it somehow?

neomilium: Is my reply clear enough to resolve to conversation?

ekohl: I do wonder if the `--help` description for the command could explain this behavior but I don't consider it blocking.

lib/modulesync.rb
neomilium: OK, I revert the current behavior, ie. skipping to whole workspace preparation when running in offline mode.
I renamed the commit, as I do need to support to prepare the workspace in offline mode for other commands.

ekohl: I was just thinking: you could still make the `prepare_workspace` accept an `operate_offline` command but simply not pass it in `manage_module` and explain in a (code) comment why it's not using the parameter. Then you could still use it the other way in your other commands. Would that make sense or is it overcomplicating things?

lib/modulesync.rb
neomilium: `msync help exec` now explain this behavior.

lib/modulesync.rb
neomilium: Checkout the updated commit, that's what I did :-)

ekohl (approved):
I think this addresses all my concerns. It's still a large patch and I could have missed things, but I'm going to trust all the tests you added. Let's try this out since I'm excited to replace my bash for loops with `msync exec`, see how that feels. Fun fact: I actually prefer doing `msync exec git pull` over trusting msync. Feels like I get more control. Certainly something I had desired for a long time.

ekohl:
Now that it's been released I'm really enjoying `msync exec`, but the one thing that bugs me is that it switches to the default branch. I had expected it to be a short hand for `for module in modules/*/* ; do ( cd $i && $command ) ; done`. Is this intentional or a bug?

ekohl:
So I opened https://github.com/voxpupuli/modulesync/pull/251 for further discussion.

PR #245, opened 2022-01-25
Add Ruby 3.1 to CI matrix and fix rubocop violations

bastelfreak:


lib/modulesync/renderer.rb
```diff
@@ -34,9 +34,7 @@ def self.render(_template, configs = {}, metadata = {})
     def self.sync(template, target_name)
       path = target_name.rpartition('/').first
```
ekohl, line 35: Unrelated, but isn't the right way `File.dirname(target_name)`? Then you can also drop the `unless path.empty?` on the next line since it's guaranteed to return a path (`.` if `target_name` is a relative path).

ekohl (approved):

ekohl:
Since we just merged a bit patch, would you mind rebasing?

lib/modulesync/renderer.rb
ekohl: Submitted as https://github.com/voxpupuli/modulesync/pull/247

PR #246, opened 2022-01-31
Fix CI

smortex:
Auto-correct rubocop offenses

### DIFF
--- a/modulesync.gemspec
+++ b/modulesync.gemspec
@@ -32,4 +32,5 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'octokit', '~>4.0'
   spec.add_runtime_dependency 'puppet-blacksmith', '>= 3.0', '< 7'
   spec.add_runtime_dependency 'thor'
+  spec.metadata['rubygems_mfa_required'] = 'true'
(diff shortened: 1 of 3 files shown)


modulesync.gemspec
```diff
@@ -32,4 +32,5 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'octokit', '~>4.0'
   spec.add_runtime_dependency 'puppet-blacksmith', '>= 3.0', '< 7'
   spec.add_runtime_dependency 'thor'
+  spec.metadata['rubygems_mfa_required'] = 'true'
```
smortex, line 35: Rubocop wants this, I do not think it is a bad idea, but maybe GitHub action will be a blocker for this for the release action?

bastelfreak: I do not think we currently support 2FA with our setup :( I will try to find some time and investigate

bastelfreak: If you disable the cop for now I can merge it if you like

smortex: @neomilium did exactly this in #244 so we can close this PR ;-)

PR #248, opened 2022-02-22
CLI: Show relevant help when using --help option on a subcommand

neomilium:
A tiny PR to (again) workaround a strange behavior of `thor`...

This one allow to use `msync subcommand --help` to show the subcommand help.

@smortex : You should like this one ;-)

### DIFF
--- a/lib/modulesync/cli/thor.rb
+++ b/lib/modulesync/cli/thor.rb
@@ -8,7 +8,19 @@ module ModuleSync
     # This class extends ::Thor class to
     # - exit with status code sets to `1` on Thor failure (e.g. missing required option)
     # - exit with status code sets to `1` when user calls `msync` (or a subcommand) without required arguments
+    # - show subcommands help using `msync subcommand --help`
     class Thor < ::Thor
+      def self.start(*args)
+        if (Thor::HELP_MAPPINGS & ARGV).any? && subcommands.none? { |command| command.start_with?(ARGV[0]) }
+          Thor::HELP_MAPPINGS.each do |cmd|
+            if (match = ARGV.delete(cmd))
+              ARGV.unshift match
+            end
+          end
+        end
+        super
+      end
+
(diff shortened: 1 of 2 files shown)


bastelfreak (approved):

lib/modulesync/cli/thor.rb
```diff
@@ -8,7 +8,19 @@ module CLI
     # This class extends ::Thor class to
     # - exit with status code sets to `1` on Thor failure (e.g. missing required option)
     # - exit with status code sets to `1` when user calls `msync` (or a subcommand) without required arguments
+    # - show subcommands help using `msync subcommand --help`
     class Thor < ::Thor
+      def self.start(*args)
+        if (Thor::HELP_MAPPINGS & ARGV).any? && subcommands.grep(/^#{ARGV[0]}/).empty?
```
ekohl, line 14: Should this use `Regexp.escape()`?
```suggestion
        if (Thor::HELP_MAPPINGS & ARGV).any? && subcommands.grep(/^#{Regexp.escape(ARGV[0])}/).empty?
```

Or perhaps even better:
```suggestion
        if (Thor::HELP_MAPPINGS & ARGV).any? && subcommands.any? { |command| command.start_with?(ARGV[0]) }
```

ekohl (commented):
This is something I've run into many times.

lib/modulesync/cli/thor.rb
neomilium: Applied (with fix: `none?` instead of `any?`)

ekohl: Oh yes, my bad. Thanks for not blindly following suggestions :)

ekohl (approved):

smortex:
> @smortex : You should like this one ;-)

I arrive after the battle but :purple_heart: :blue_heart:  :green_heart: :yellow_heart: :orange_heart: :heart:

Issue #192, opened 2020-08-16
Target branch `.sync.yml` not taken into account on branch update (--force)


bittner:
When updating an existing branch, `msync update --force` seems to consider the values in the `.sync.yml` file of the _existing_ branch (meant to be updated), instead of the `.sync.yml` file of the _target_ branch. This might be a design fault.

**Side note:** I did this on a GitLab merge request, but the behavior is likely the same even w/o `--pr`

## Example update flow

1. I run `msync update --pr`, which creates an update (new branch, optional merge request)
1. I figure out that the change was wrong on a target module, due to a value I forgot to update in `.sync.yml`; hence, I update `.sync.yml` (in target module, in target branch, e.g. `master`).
1. I run `msync update --pr --force` to update the already existing branch (created earlier)

Now, I would expect the updated branch to consider the values in `.sync.yml` of the target branch (i.e. `master` in our example). But the changes are based on the values of `.sync.yml` from the existing, earlier created branch.

I need to delete the branch created by `msync update` and run ModuleSync afresh.  :worried: 

## Recreate target branch? (instead of "update")

The above behavior is probably not intended. The branch was originally created off the target branch, which then naturally has a `.sync.yml` file that matches on both branches.

When a branch is updated, especially with `--force`, then the branch should really, technically, be created off that target branch again and force-pushed over the branch created by ModuleSync earlier. Correct?

neomilium:
I had similar troubles using `msync` but according to the help, `--force` is not the option for this expected behavior:
```shell-session
$ bundle exec msync help update
Usage:
  msync update

Options:
[...]
      [--force]                                      # Force push amended commit
[...]
```

IMHO, in `msync update` we lack of a `--reset-hard` option, in order to hard-reset on the specified branch before updating files then push again the result.
That's why I add it in #199 at 5942b00593191d005a5f2f947b195761a212a2fe.

More precisely, I dont catch the usecase when you want to amend an existing commit then push if we had an hard-reset feature.
AFAICS, amending a commit is more dangerous for mass manipulation, as you dont ever know is the previously commit is what you expected.

neomilium:
In order to illustrate your issue (and test the new framework available in #202), I wrote what I understand of your issue as a behavior test:

```gherkins
Feature: Update module with a .sync.yml file

  @wip
  @announce-output
  Scenario: Updating a module with a .sync.yml file, then update it upstream
    Given a basic setup with a puppet module "puppet-test" from "fakenamespace"
[...]

PR #251, opened 2022-03-09
exec command: Only switch branch when cloning

ekohl:
When using msync exec, I was surprised to learn it switched branches. My immediate impression was that it was simply a wrapper that executes commands on checkouts.

This changes the commend to only switch branch when cloning the repository or when it's explicitly passed via the --branch option.

Right now there are no tests (I did verify it manually) because I first want to figure out if this is a good idea. @smortex would you mind having a look?

### DIFF
--- a/lib/modulesync.rb
+++ b/lib/modulesync.rb
@@ -199,8 +199,12 @@ module ModuleSync # rubocop:disable Metrics/ModuleLength
     managed_modules.each do |puppet_module|
       $stdout.puts "#{puppet_module.given_name}:"
 
-      puppet_module.repository.clone unless puppet_module.repository.cloned?
-      puppet_module.repository.switch branch: @options[:branch]
+      if puppet_module.repository.cloned?
+        puppet_module.repository.switch branch: @options[:branch] if @options[:branch]
+      else
+        puppet_module.repository.clone
+        puppet_module.repository.switch branch: @options[:branch]
+      end
 
       command_args = cli_options[:command_args]
       local_script = File.expand_path command_args[0]


bittner:
Heads-up: Please make sure you don't break the behavior fixed via #192 and #242.

ekohl:
Note that this is the `exec` command. I'm not sure that even takes `.sync.yml` into account. I thought that file is only needed for rendering templates.

smortex:
:thinking: quite unexpected as far as I think about it, yes. I had a similar issue but not exactly the same I think and @neomilium demonstrated me I was inconsistent and changing branch was better… Maybe it's not applicable here and maybe it's not applicable anymore anywhere after the recent refactoring, so let's let him shine and enlighten us: is this expected :grin: ?

neomilium:
`exec` is a new feature without explicit behavior expected, but the most important thing to me during development was to keep some kind of _idempotency_ / _reproductibility_.
ie. Running the same command at different times and different contexts should produce the same result

My first idea was to set `--branch` as required option, then I add some kind of flexibility to allow "default" branch to be selected when branch wasn't supplied... I agree with you: it could be unexpected to run it without `--branch` option but do switch branch.

I'm not so comfortable with the supplied patch as it produce a different behavior if cloned? or not.

Maybe it could be an idea, to implement something like this:
 - check if each repo have been cloned before running `exec`
 - raise unless all repo are cloned
 - turn `--branch` into required option unless `--keep-current-checkout` is set

@ekohl It seems we use a different workflow to use `msync`, if you have some contextualized use case where its relevant to keep current checkout, I do want to ear it.

Nevertheless, I'm OK to merge this if main contributors too, maybe some behavior tests could help to keep the feature when the expected behavior were found.

ekohl:
> @ekohl It seems we use a different workflow to use `msync`, if you have some contextualized use case where its relevant to keep current checkout, I do want to ear it.

I'll share my previous example. I started work on something that wasn't ready, but I wanted to see the impact. So I started work on a change:
```console
$ msync update --noop -b my-change
```

Then wanted to figure out what changed, so I ran
```console
$ msync exec -- git diff
```

This led to unexpected results.

Other actions I can imagine are `msync exec -- bundle update && bundle exec rubocop -a && git commit -am 'Rubocop autofix'`.

Maybe my core issue is that I like having control over branches. Perhaps similar to `msync update`, it should have `--offline` to not perform any git operations. I've already guaranteed that everything is in order with `msync update` in a previous step.

> Nevertheless, I'm OK to merge this if main contributors too, maybe some behavior tests could help to keep the feature when the expected behavior were found.

Any thoughts from @bastelfreak or @alexjfisher? I think you're semi-regular users.

PR #252, opened 2022-04-26
Handle Ruby 3.1 ERB trim_mode deprecation

ekohl:
In Ruby 3.1 using `trim_mode` must be passed as a keyword argument or a deprecation warning is shown. However, that is only possible on Ruby 2.7 or newer.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -12,6 +12,9 @@ AllCops:
     - 'Gemfile'
     - 'Rakefile'
 
+Style/FetchEnvVar:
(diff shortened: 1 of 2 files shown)


ekohl:
The Rubocop failure is unrelated, but I think we should disable https://docs.rubocop.org/rubocop/cops_style.html#stylefetchenvvar because IMHO it's a bad case. Especially since this gem is already handling it anyway.

bastelfreak (approved):

PR #254, opened 2022-09-14
Expose namespace in metadata

ekohl:
While the namespace is already in @configs, it feels more natural in metadata.

### DIFF
--- a/features/update.feature
+++ b/features/update.feature
@@ -717,6 +717,7 @@ Feature: update
     And a file named "moduleroot/test.erb" with:
       """
       module: <%= @metadata[:module_name] %>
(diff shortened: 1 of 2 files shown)


bastelfreak (approved):
besides the rubocop complaints, I'm fine with the change.

ekohl:
> besides the rubocop complaints, I'm fine with the change.

Those happen in master too: https://github.com/voxpupuli/modulesync/pull/255

ekohl:
Can we drop codecov? The change is covered, but through cucumber rather than unit tests.

bastelfreak (approved):
looks good to me. if you like to remove codecov I'm fine with that.

PR #260, opened 2023-03-09
Fix compatibility with latest `ruby-git`

alexjfisher:
Since `ruby-git` `1.16.0` it is no longer possible to parse the output of `branch.full` to determine the default branch. (It now only returns `"remotes/origin/HEAD"` where previously it returned `"remotes/origin/HEAD -> origin/master"`)

Fortunately, since `1.17.0` (which was only released a couple of days after `1.16.0`), there is a `Git.default_branch` method we can use instead.

This commit uses the new method if available.  In the future, we can probably drop the old code and require `ruby-git` `>= 1.17.0`, but since the project seems to currently be undergoing quite a bit of development and we might encounter other issues, lets keep it in for now.

Fixes #259

PR #261, opened 2023-03-09
dependabot: check for github actions and gems

bastelfreak:


bastelfreak:
the rubocop failures are not related to the change and alex fixed them in https://github.com/voxpupuli/modulesync/pull/260

PR #265, opened 2023-04-14
Update to latest RuboCop 1.28.2

bastelfreak:


bastelfreak:
that's the newest version that still supports Ruby 2.5

PR #282, opened 2024-05-17
pin thor to 1.3.0

bastelfreak:
With thor 1.3.1 we've two failing tests:

```
Failing Scenarios:
cucumber features/execute.feature:36 # Scenario: Show fail-fast default value in help
cucumber features/execute.feature:44 # Scenario: Override fail-fast default value using config file
```

as a workaround to get CI green again we'r pinning to 1.3.0. I think the bug is fixed in https://github.com/rails/thor/pull/878 but not yet released.

### DIFF
--- a/modulesync.gemspec
+++ b/modulesync.gemspec
@@ -27,5 +27,5 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'gitlab', '~>4.0'
   spec.add_runtime_dependency 'octokit', '>=4', '<9'
   spec.add_runtime_dependency 'puppet-blacksmith', '>= 3.0', '< 8'
-  spec.add_runtime_dependency 'thor'
+  spec.add_runtime_dependency 'thor', '1.3.0'
 end


modulesync.gemspec
```diff
@@ -27,5 +27,5 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'gitlab', '~>4.0'
   spec.add_runtime_dependency 'octokit', '>=4', '<9'
   spec.add_runtime_dependency 'puppet-blacksmith', '>= 3.0', '< 8'
-  spec.add_runtime_dependency 'thor'
+  spec.add_runtime_dependency 'thor', '1.3.0'
```
ekohl, line 30: Should it have pinned to NOT 1.3.0?

bastelfreak: I think that's correct? 1.3.1 has a regression and https://github.com/rails/thor/pull/878 contains a potential fix.

ekohl: I meant to pin to `!= 1.3.1`. I shouldn't send comments as I'm changing trains using the mobile app. Never ends well

bastelfreak: ah ok. I considered that, but I don't know if the fix will be in 1.3.2 or later. So I think the safer way is pinning to 1.3.0 and dependabot will inform us for new releases.

ekohl: The reason I mention it is because dependabot also informed us about 1.3.1, but yes, this stuff is always annoying

PR #301, opened 2025-05-21
CI: Generate ruby matrix dynamically

bastelfreak:
contains https://github.com/voxpupuli/modulesync/pull/300


ekohl (approved):
:+1: assuming tests are green.

PR #303, opened 2025-05-21
thor: Allow 1.2 to stay compatible with ancient facter

bastelfreak:
For people bundling various tools from the ecosystem in a container, lets say https://github.com/voxpupuli/container-voxbox, we want to / have to stay compatible with facter for now. The
ancient/dead/unmaintained Perforce facter gem depends on thor < 1.3.


TheMeier (approved):
:facepalm:

PR #311, opened 2025-07-25
Allow YAML aliases in configuration

traylenator:
Currently with our configuration we use aliases on the config_defaults.yml.

https://gitlab.cern.ch/ai/it-puppet-modulesync-configs/-/blob/master/config_defaults.yml


lib/modulesync/util.rb
```diff
@@ -12,7 +12,7 @@ def self.symbolize_keys(hash)
 
     def self.parse_config(config_file)
       if File.exist?(config_file)
-        YAML.load_file(config_file) || {}
+        YAML.load_file(config_file, aliases: true) || {}
```
smortex, line 15: Older versions of ruby do not have support for this keyword.  Given we will drop Puppet 7 ([discussion](https://github.com/voxpupuli/community-triage/issues/60)), testing with these old ruby versions do not add value and I would rather remove them.

smortex (approved):

PR #350, opened 2026-07-09
fix: add guess_endpoint_from for github, similar to gitlab

rwaffen:
if you forget to export GITHUB_BASE_URL it now should return that for github it is https://api.github.com and do not fail anymore.

The priority is:
- Per-repository base_url in managed_modules.yml
- GITHUB_BASE_URL
- Automatic detection via guess_endpoint_from


sebastianrakel (approved):

rwaffen:
ci!? why fail? 😢

PR #353, opened 2026-07-09
ci: add labeler workflow

rwaffen:


bastelfreak (approved):

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,18 @@
+---
+name: 🏷️ Pull Request Labeler
+
+on:
+  - pull_request_target
+
+permissions:
```
bastelfreak: you can set the permissions to zero here. I think it's `permissions: {}`?

rwaffen: we can try that! thought we have to have to do it this way... because i saw you doing it everywhere :D

PR #368, opened 2026-07-24
feat: add support for cryptographic signing and signoff options in commits and tags

rwaffen:


rwaffen:
we need atleast the signoff option because you still cannot set it via gitconfig. and for OV we need to signoff commits. otherwise PRs from pccibot will not be mergeable

README.md
```diff
@@ -166,6 +166,17 @@ Not required for gerrit.
 msync update --amend --force
 ```
 
+Use `--sign` to cryptographically sign commits and tags with Git's configured signing key and backend.
```
bastelfreak, line 169: the `--amend` options is also new? Could you document it as well?

rwaffen: is it? its not from me, but i can document it, sure.

bastelfreak (approved):

